Issue #652, opened 2024-07-23
Update the handling of interactions with the Deposit/Withdrawal form


michalinacienciala:
We should update the Deposit/Withdrawal forms to behave in a way described in this doc: https://coda.io/d/Acre_d3Vs1FswTQA/Deposit-Withdraw-form-interactions_suN-q#_ludJH.
This doc is a result of conversations [here](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=15225-12057&t=ZMzo62UrftNOz9Th-4), [here](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=15236-12487&t=ZMzo62UrftNOz9Th-4) and [here](https://discord.com/channels/1079490158639976609/1264882672338210918/1264882674611261453)

Issue #637, opened 2024-07-22
`Required` error should not be shown before user presses Deposit/Withdraw button


michalinacienciala:
As per https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=15225-12057&t=jgRcGrbRWYOUj94I-0:

> If I click inside the input field and then outside, I receive this error, which should only be displayed if the user presses the Deposit button before adding an amount.

![image](https://github.com/user-attachments/assets/f94a64d7-80e3-463c-bd2c-bebe03511869)

after clicking 'Deposit' it should look like this:

![image](https://github.com/user-attachments/assets/2341b4f3-4a85-4bf4-8846-2ea4d5f66191)

Issue #630, opened 2024-07-19
Improve formatting of the 'Amount' input


michalinacienciala:
As per https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8?node-id=14880-13400#876189443:

Currently:
![image](https://github.com/user-attachments/assets/714af311-0ae9-4f49-a44d-a52b38fd2ede)

Expected:
![image](https://github.com/user-attachments/assets/467f7ee9-e09c-4c0e-a639-4c670f75ffe9)

![image](https://github.com/user-attachments/assets/4b048368-7260-4f56-b0d6-7181129953a0)

Issue #629, opened 2024-07-19
Missing focus state on the 'Amount' input field


michalinacienciala:
As per https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8?node-id=14880-13398#876185850 and https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=15225-12057&t=jgRcGrbRWYOUj94I-0:

![image](https://github.com/user-attachments/assets/3df5ab64-995a-4e25-80dd-23a35dd792f9)

Applies to both the deposit and withdrawal form.

PR #687, opened 2024-08-01
dApp: Deposit/Withdrawal form improvements - visuals

kpyszkowski:
Closes: #629
Closes: #652
Closes: #637
Closes: #570 
Closes: #630

### Changes:
- Adjusted styles: spacings, button styles, input
- Form input is automatically focused as the form appears
- Added stake form input placeholder
- Added withdrawal form input placeholder
- Disabled submit button when amount is invalid
- Removed fiat-converted amount displayed below the input
- Displayed validation errors as `Alert` component
- Adjusted validation strategy
- Updated input displayed value formatting function to use `bigIntToUserAmount` instead of `fixedPointNumberToString`:
  - The `fixedPointNumberToString` aggressively trims trailing zero values conflicting with the way `NumericFormat` component does it. The `bigIntToUserAmount` seems to fit better here
- Updated copies
- Conditionally styled balance amount for invalid state
- Added suffix currency symbol value

https://github.com/user-attachments/assets/17c203ae-0b6d-45e6-a47b-f506609d24b9


kkosiorowska (dismissed):
LGTM, ready for testing 🚀

michalinacienciala:
On the Withdraw form the `Wallet balance` label should be replaced by `Acre balance`, as per [this](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8?node-id=15225-12057#881137790) comment.

michalinacienciala:
The input field should show a grey 'BTC' text after the amount entered by the user.

Design:
image

Actual:
image

michalinacienciala:
The amount in the input field should not change the color to red when there's a failure of the validation.

Design:
image

Actual:
image

michalinacienciala:
The value inside the input field should be `0.015 BTC`, not `0.0150 BTC` (and for withdrawal: `0.01 BTC` instead of `0.0100 BTC`):

Actual:
image

[Design](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=15225-12057&t=7Iw13UFemvH5INqx-0):
image

michalinacienciala:
As per [this](https://coda.io/d/Acre_d3Vs1FswTQA/Deposit-Withdraw-form-interactions_suJjFN-q#_luSfMc3A) doc and [this](https://discord.com/channels/1079490158639976609/1240933341235253310/1273949269224853525) Discord discussion, we shouldn't show errors relating to the values entered in the amount input until user clicks Deposit/Withdraw. Right now we're validating the amount dynamically.

michalinacienciala:
Also, just making sure - the handling of [the insufficient balance to handle fee](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=14697-5787&t=JtGEYez3vVEh8Oc1-0) is not in scope of this PR, right?

kpyszkowski:
@michalinacienciala 

> The input field should show a grey 'BTC' text after the amount entered by the user.

In designs to looks like the value is mapped over placeholders. It requires the entire component to be rebuilt because we don't have such functionality and it's not how native inputs work. I suppose it would be too much time consuming to us. I'd suggest to skip it.

> The amount in the input field should not change the color to red when there's a failure of the validation.

Ref commit: 1554607c0f27691b4fb510871d9ad391c8368d4f

> The value inside the input field should be 0.015 BTC, not 0.0150 BTC (and for withdrawal: 0.01 BTC instead of 0.0100 BTC)

Ref commit: 278c5432c189befb2450ac54f964d93c765c413f

> As per [this](https://coda.io/d/Acre_d3Vs1FswTQA/Deposit-Withdraw-form-interactions_suJjFN-q#_luSfMc3A) doc and [this](https://discord.com/channels/1079490158639976609/1240933341235253310/1273949269224853525) Discord discussion, we shouldn't show errors relating to the values entered in the amount input until user clicks Deposit/Withdraw. Right now we're validating the amount dynamically.
Also, the Deposit/Withdraw button should be disabled only freshly after user clicked on it and the validation failed. Afer that any change to the amount field should enable the button.

Functionalities you describe are impossible to implement with Formik - the library we use. It's impossible to track the state before first submit. About the form behavior that's all we can do.

michalinacienciala:
> Also, just making sure - the handling of [the insufficient balance to handle fee](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=14697-5787&t=JtGEYez3vVEh8Oc1-0) is not in scope of this PR, right?

@kpyszkowski I think you missed this one

michalinacienciala:
Regarding the two items deemed as impossible / hard to implement - I asked @SorinQ about the approach we should take here. Anyway, we can probably close this PR without those changes and handle them in separate tasks (as soon as we resolve all other issues in the current PR).

dapp/src/components/shared/Form/FormTokenBalanceInput.tsx
```diff
@@ -28,14 +29,28 @@ export function FormTokenBalanceInput({
...
+
+      if (!meta.error) {
+        return
+      }
+
+      helpers.setError(undefined)
+    },
+    [field, meta, helpers],
+  )
+
```
kkosiorowska, line 44: Non-blocking comment

I wonder why not just do it in `setAmount`. Note that in this function we also set `touched` to `true`. I would suggest keeping such changes together to maintain some consistency. So that later we don't wonder whether further changes should be added to `setAmount` or `handleChange`. What do you think about it?

https://github.com/thesis/acre/blob/dc156f5a7f02142c1f80627267d14a26e5c99b30/dapp/src/components/shared/Form/FormTokenBalanceInput.tsx#L19-L20

We could then simplify the code to

```
  const setAmount = useCallback(
    (value?: bigint) => {
      if (!meta.touched) logPromiseFailure(helpers.setTouched(true))
      if (meta.error) helpers.setError(undefined)

      logPromiseFailure(helpers.setValue(value))
    },
    [helpers, meta.error, meta.touched],
  )
```

dapp/src/components/shared/Form/FormTokenBalanceInput.tsx
```diff
@@ -28,14 +29,28 @@ export function FormTokenBalanceInput({
...
 
+  const handleChange = useCallback<ChangeEventHandler>(
+    (event) => {
+      field?.onChange(event)
+
+      if (!meta.error) {
+        return
+      }
+
+      helpers.setError(undefined)
```
kkosiorowska, line 40:
```suggestion
if (meta.error) helpers.setError(undefined)
```

dapp/src/components/shared/TokenBalanceInput/index.tsx
```diff
@@ -118,27 +126,38 @@ export default function TokenBalanceInput({
...
-  const { decimals } = getCurrencyByType(currency)
+  const { decimals, symbol } = getCurrencyByType(currency)
 
   const handleValueChange = (value: string) => {
     valueRef.current = value ? userAmountToBigInt(value, decimals) : undefined
   }
 
+  const handleChange: ChangeEventHandler<HTMLInputElement> = (event) => {
+    inputProps.onChange?.(event)
+    setAmount(valueRef?.current)
```
kkosiorowska, line 143: I was curious if `inputProps.onChange?.(event)` would affect the setting of the amount value. So I switched their order to tests.

After entering the value 5 , `inputProps.onChange?.(event)` changed the value to 0.0005. This is probably because our field is of non-standard bigint type and the event passes a string value to the `onChange` method. So I have some worries about calling `onChange` here. I think we should not allow the `onChange` function to be passed to `TokenBalanceInput`. I'm curious what you think.

https://github.com/user-attachments/assets/f3958cff-6597-41cc-ba8e-52ac69694610

kkosiorowska (commented):
Overall it looks good. I left a few small questions. Let me know what you think.

Additionally, I will wait for the merge until tomorrow in case Michalina finds a moment to test. If not, we will do the merge tomorrow.

dapp/src/components/shared/Form/FormTokenBalanceInput.tsx
kpyszkowski: That's one of the solutions. Have to admit your suggestion is cleaner. Thanks
Ref commit: ea855814

dapp/src/components/shared/Form/FormTokenBalanceInput.tsx
kpyszkowski: Used in ea8558141dd662758dbf9a90673a2ce396042a26

dapp/src/components/shared/TokenBalanceInput/index.tsx
kpyszkowski: After changes from ea85581 the `onChange` handler call is redundant.
The explicit `onChange` call here was to ensure the change event is not lost in event propagation when I introduced the proxy event handler [here](https://github.com/thesis/acre/pull/687/commits/a72361a84776816593208a56d93e255c7640f83f#diff-69b290384051dd99cdfc367978535151273cc0eb90dcbf201bd20d03ee22b43dR32). Since I reset errors in `setAmount` callback  it's no longer valid.

kkosiorowska (approved):
🚀🚀🚀

dapp/src/components/shared/NumberFormatInput/index.tsx
```diff
@@ -29,17 +35,35 @@ const NumberFormatInput = React.forwardRef<
...
+  const handleLengthValidation = (values: NumberFormatValues) => {
+    const { value } = values
+    if (!value || !maxIntegerLength || !maxDecimalLength) return true
+
+    const [integer, decimal] = value.split(".")
+    const isValidIntegerLength = integer && integer.length <= maxIntegerLength
+    const isValidDecimalLength = decimal && decimal.length <= maxDecimalLength
+
+    return decimal ? isValidDecimalLength : isValidIntegerLength
+  }
```
kkosiorowska, line 56: When we get rid of `maxDecimalLength` we can simplify it a bit. We want to check the integer part from the value. So we can use the `parseInt` method. What do you think about it?

I noticed two more things:

1. The `!maxIntegerLength` return `true` when `maxIntegerLength=0`. I think in such a case we would want to move on instead of exiting the function.
2. The user can enter the value `.` to input. That's why I would check if `floatValue` is defined to correctly convert the `value` to integer.

```suggestion
  const handleLengthValidation = (values: NumberFormatValues) => {
    const { value, floatValue } = values

    if (floatValue === undefined || maxIntegerLength === undefined) return true

    const convertedValue = parseInt(value, 10)
    const valueLength = convertedValue.toString().length

    return valueLength <= maxIntegerLength
  }
```

dapp/src/components/shared/TokenBalanceInput/index.tsx
```diff
@@ -152,18 +170,19 @@ export default function TokenBalanceInput({
...
             handleValueChange(values.value)
           }
-          onChange={() => {
-            setAmount(valueRef?.current)
-          }}
+          onChange={handleChange}
           decimalScale={decimals}
           allowNegative={false}
+          maxIntegerLength={10}
+          maxDecimalLength={decimals}
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 880bbd9

dapp/src/components/shared/NumberFormatInput/index.tsx
```diff
@@ -29,17 +35,35 @@ const NumberFormatInput = React.forwardRef<
...
+    ...restProps
+  } = props
+
+  const handleLengthValidation = (values: NumberFormatValues) => {
+    const { value } = values
+    if (!value || !maxIntegerLength || !maxDecimalLength) return true
+
+    const [integer, decimal] = value.split(".")
+    const isValidIntegerLength = integer && integer.length <= maxIntegerLength
+    const isValidDecimalLength = decimal && decimal.length <= maxDecimalLength
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 880bbd9

dapp/src/components/shared/NumberFormatInput/index.tsx
```diff
@@ -29,17 +35,35 @@ const NumberFormatInput = React.forwardRef<
...
+
+    return decimal ? isValidDecimalLength : isValidIntegerLength
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 880bbd9

dapp/src/components/shared/TokenBalanceInput/index.tsx
```diff
@@ -152,18 +170,19 @@ export default function TokenBalanceInput({
...
           onValueChange={(values: NumberFormatInputValues) =>
             handleValueChange(values.value)
           }
-          onChange={() => {
-            setAmount(valueRef?.current)
-          }}
+          onChange={handleChange}
           decimalScale={decimals}
           allowNegative={false}
+          maxIntegerLength={10}
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 880bbd9

dapp/src/components/shared/NumberFormatInput/index.tsx
kpyszkowski: Ref commit: 880bbd9

kkosiorowska (approved):
Confirmed works great! 🚀

Screenshot 2024-08-21 at 11 00 28

PR #708, opened 2024-08-20
Update the dApp to work in Ledger Live

kkosiorowska:
This is an integration PR that adds the necessary changes for the dapp to run on Ledger Live.

Refs: #730 #742 #763 #759, #760, #797, #780, https://github.com/thesis/orangekit/pull/143, https://github.com/thesis/orangekit/pull/144

### Important note
Changes can only be tested in a Ledger Live app built from source. See the [Ledger Live repo](https://github.com/LedgerHQ/ledger-live), how to build from source.


dapp/ledger-manifest-testnet-preview.json
```diff
@@ -0,0 +1,31 @@
+{
+  "id": "acre-testnet",
+  "name": "Acre [testnet - preview]",
+  "url": "https://deploy-preview-708--acre-dapp-testnet.netlify.app/?embed=ledger-live",
```
nkuba, line 4: We configured netlify branch deployments for this feature branch, so we can use `https://ledger-live-updates--acre-dapp-testnet.netlify.app/?embed=ledger-live` here.

dapp/src/hooks/useReferral.ts
```diff
@@ -17,10 +18,17 @@ export default function useReferral(): UseReferralReturn {
...
   )
   const { openModal } = useModal()
+  const { isEmbed } = useIsEmbed()
 
   const detectReferral = useCallback(() => {
     const param = referralProgram.getReferralFromURL()
 
+    if (isEmbed) {
+      // TODO: Set correct referral for embedded dApp
+      setReferral(0)
```
nkuba, line 28: The `embed` parameter is not boolean. We will use it to pass the partner integration ID what will have to map to a referral ID. For Ledger Live we will have `embed=ledger-live` and we need to map it to the referral ID assigned for Ledger Live, which is number `2083` (source: [Coda page](https://coda.io/d/Acre_d3Vs1FswTQA/Referrals_sudl9wjh#_lubtctD5))

dapp/src/hooks/useLastUsedBtcAddress.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback } from "react"
+import useLocalStorage from "./useLocalStorage"
+
+export const LAST_USED_BTC_ADDRESS_KEY = "lastUsedBtcAddress"
```
nkuba, line 4: Can we keep the `LAST_USED_BTC_ADDRESS_KEY` unexported? We could define helper functions to read and write the last used btc address in the storage. Then we could use the read function in `dapp/src/wagmiConfig.ts`.

Originally posted in https://github.com/thesis/acre/pull/742#discussion_r1782879026

dapp/package.json
```diff
@@ -13,14 +13,18 @@
     "lint:js:fix": "eslint . --fix",
     "lint:config": "prettier -c '**/*.@(json|yaml|toml)'",
     "lint:config:fix": "prettier -w '**/*.@(json|yaml|toml)'",
-    "test": "vitest"
+    "test": "vitest",
+    "prepare:dev": "mustache config/development.json ledger-manifest-template.json > ledger-manifest-development.json",
+    "prepare:testnet": "mustache config/testnet.json ledger-manifest-template.json > ledger-manifest-testnet.json",
+    "prepare:mainnet": "mustache config/mainnet.json ledger-manifest-template.json > ledger-manifest-mainnet.json",
+    "prepare:all-env": "pnpm prepare:dev && pnpm prepare:testnet && pnpm prepare:mainnet"
```
nkuba, line 20: We could simplify the settings by combining the scripts in one:
```suggestion
    "generate:manifests": "for env in development testnet mainnet; do mustache config/$env.json ledger-manifest-template.json > ledger-manifest-$env.json; done"
```

Additionally we could include the generation in the build step:
```json
"build": "npm run generate:manifests && tsc && vite build",
```

dapp/config/testnet.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "testnet",
+  "name": "Acre [testnet]",
+  "url": "https://dapp.test.acre.fi/"
```
nkuba, line 4:
```suggestion
  "url": "https://stake.test.acre.fi/"
```

dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
```diff
@@ -127,6 +127,8 @@ export default function ConnectWalletButton({
       onError: (error: OrangeKitError) => {
         const errorData = orangeKit.parseOrangeKitConnectionError(error)
         setConnectionError(errorData)
+        if (connector.id === "orangekit-ledger-live")
+          logPromiseFailure(connector.disconnect())
```
nkuba, line 131: We could avoid this statement by handling the problem in the Orange Kit Ledger Live Provider inside the connect function. In the for loop, if `getAddress` fails with `UnsupportedBitcoinAddressError`, we could catch the error and continue the loop instead of failing the execution.

r-czajkowski: e50d86d

dapp/src/components/ConnectWalletModal/index.tsx
```diff
@@ -1,22 +1,25 @@
-import React, { useState } from "react"
+import React, { useEffect, useState } from "react"
 import { ModalBody, ModalHeader, ModalCloseButton } from "@chakra-ui/react"
-import { useConnectors, useWalletConnectionError } from "#/hooks"
+import { useConnectors, useIsEmbed, useWalletConnectionError } from "#/hooks"
 import { OrangeKitConnector, BaseModalProps, OnSuccessCallback } from "#/types"
-import { wallets } from "#/constants"
+// import { wallets } from "#/constants"
```
kkosiorowska, line 5:
```suggestion
```

dapp/src/hooks/orangeKit/useBitcoinProvider.ts
```diff
@@ -8,7 +8,8 @@ export function useBitcoinProvider(): UseBitcoinProviderReturn {
   const connector = useConnector()
 
   return useMemo(() => {
-    if (!connector) return undefined
+    if (!connector || typeof connector.getBitcoinProvider !== "function")
+      return undefined
```
kkosiorowska, line 12: I'm wondering why we added this condition here. This issue is also addressed [here](https://github.com/thesis/acre/pull/771).  Hook `useConnector` returns a connector from OrangeKit which should have an implemented method `getBitcoinProvider`.  Therefore, it seems to me that this is not right place for the fix.

dapp/src/hooks/useTriggerConnectWalletModal.ts
```diff
@@ -0,0 +1,22 @@
...
+  const { isEmbed } = useIsEmbed()
+  const { openModal, closeModal } = useModal()
+  const isMounted = useRef(false)
+
+  useEffect(() => {
+    if (!isMounted.current && isEmbed && !isSignedMessage) {
+      isMounted.current = true
+      openModal(MODAL_TYPES.CONNECT_WALLET, {
+        withCloseButton: true,
+        closeOnEsc: false,
```
kkosiorowska, line 18: I think we should also set  `true` here.
```suggestion
        closeOnEsc: true,
```

kkosiorowska (commented):
Overall it looks good, I left non-blocking comments.

dapp/src/hooks/orangeKit/useBitcoinProvider.ts
r-czajkowski: Wagmi sometimes returns connector w/o functions I'm not sure why - I noticed this in logs that we got an uncaught error and fixed it.

dapp/src/components/ConnectWalletModal/index.tsx
r-czajkowski: 60b474d

dapp/src/hooks/useTriggerConnectWalletModal.ts
r-czajkowski: 2db8296

dapp/src/components/ConnectWalletModal/index.tsx
```diff
@@ -1,37 +1,61 @@
...
   onSuccess?: OnSuccessCallback
+  isReconnecting?: boolean
 } & BaseModalProps) {
+  const { isEmbed } = useIsEmbed()
+  const { onDisconnect } = useWallet()
   const connectors = useConnectors()
   const enabledConnectors = connectors.map((connector) => ({
     ...connector,
-    isDisabled: !wallets.SUPPORTED_WALLET_IDS.includes(connector.id),
+    isDisabled: false,
```
r-czajkowski, line 28: @kkosiorowska do we still want to be able to enable/disable wallets? I also wanted to point out that we can't control the `Unisat` wallet via environment variables.

dapp/src/wagmiConfig.ts
```diff
@@ -24,12 +28,32 @@ const transports = chains.reduce(
...
+const defaultConnectors = [
   orangeKitOKXConnector(),
   orangeKitUnisatConnector(),
   orangeKitXverseConnector(),
-] as unknown as CreateConnectorFn[]
+]
+
+const connectors = (createEmbedConnectorFn !== undefined
+  ? [createEmbedConnectorFn()]
+  : defaultConnectors) as unknown as CreateConnectorFn[]
```
r-czajkowski, line 56: We need an update here - I can use the dapp w/o access code when passing any value via URL  eg. `http://localhost:5173/?embed=test`.

dapp/package.json
```diff
@@ -13,14 +13,17 @@
...
+    "test": "vitest --reporter=verbose",
+    "generate:manifests": "for env in development testnet mainnet; do mustache config/$env.json ./manifests/ledger-live/ledger-manifest-template.json > ./manifests/ledger-live/ledger-live-manifest-$env.json; done"
   },
   "dependencies": {
     "@acre-btc/sdk": "workspace:*",
     "@chakra-ui/react": "^2.8.2",
     "@emotion/react": "^11.11.1",
     "@emotion/styled": "^11.11.0",
-    "@orangekit/react": "1.0.0-beta.32",
+    "@ledgerhq/wallet-api-acre-module": "0.1.0-nightly.0",
```
r-czajkowski, line 24: I think we can use `0.1.0` version.

dapp/src/components/ConnectWalletModal/index.tsx
kkosiorowska: >@kkosiorowska do we still want to be able to enable/disable wallets? I 

I think it is no longer necessary. It would also be good to get rid of feature flags that are no longer needed.

dapp/src/theme/utils/zIndices.ts
```diff
@@ -1,5 +1,9 @@
 export const zIndices = {
-  sidebar: 1450,
+  sidebar: 1400,
```
kkosiorowska, line 2: I just noticed this but the buttons on the sidebar don't work. This is probably because we should not change the `z-index` for the sidebar. I think we just need to back to the previous value? 🤔


https://github.com/user-attachments/assets/3f4d841d-9d35-4cf9-9aad-6f32e31fabf9

dapp/src/utils/orangekit/ledger-live/bitcoin-provider.ts
```diff
@@ -0,0 +1,323 @@
...
+      const err = error as Error
+
+      // We want to throw the same error for all connectors when user rejects
+      // the request.
+      if (
+        err?.message === "UserRefusedOnDevice" ||
+        err?.message === "Canceled by user"
+      ) {
+        throw new UserRejectedRequestError(err)
+      }
```
kkosiorowska, line 43: If we return the same error for all connectors, do we still need to use the `ledgerLive.didUserRejectRequest` call ?

https://github.com/thesis/acre/blob/0d137f162470c1dc67e5b8a7005a3168d223ca2b/dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx#L54-L66

kkosiorowska (commented):
Overall, it looks good. I left some minor comments.

I get the following error on starting the app. Can we fix this somehow?
Screenshot 2024-11-05 at 08 36 39

r-czajkowski:
> Overall, it looks good. I left some minor comments.
> 
> I get the following error on starting the app. Can we fix this somehow? Screenshot 2024-11-05 at 08 36 39

Yep, looks like it's an issue on the orangkit side. We need to add try catch here https://github.com/thesis/orangekit/blob/main/typescript/react/src/wallet/connector.ts#L151-L155. I'll open a PR in orangekit, but let's not block this PR - we'll just bump `@orangekit/react` package to fix this bug.

r-czajkowski:
> > Overall, it looks good. I left some minor comments.
> > I get the following error on starting the app. Can we fix this somehow? Screenshot 2024-11-05 at 08 36 39
> 
> Yep, looks like it's an issue on the orangkit side. We need to add try catch here https://github.com/thesis/orangekit/blob/main/typescript/react/src/wallet/connector.ts#L151-L155. I'll open a PR in orangekit, but let's not block this PR - we'll just bump `@orangekit/react` package to fix this bug.

@kkosiorowska see https://github.com/thesis/orangekit/pull/146

dapp/src/pages/DashboardPage/PositionDetails.tsx
```diff
@@ -84,7 +81,7 @@ export default function PositionDetails() {
         </UserDataSkeleton>
       </VStack>
 
-      <HStack w="full" justify="center" spacing={2}>
+      <HStack w="full" justify="start" flexWrap="wrap" spacing={2}>
         <UserDataSkeleton>
           <Button {...buttonStyles} onClick={openDepositModal}>
             {activitiesCount === 0 ? "Deposit" : "Deposit more"}
```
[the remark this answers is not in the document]
r-czajkowski: aeb1cf4

dapp/src/theme/utils/zIndices.ts
kpyszkowski: There is an issue with stacking contexts here. Currently Sidebar has lower stacking context so it remains "under" the modal - not clickable. The `1400` value of `z-index` might work here if modal and sidebar nodes were in the same root node. Reverting the value to `1450` resolves the issue but doesn't fix the issue with header. Now the header is inoperable when the modal is open eg. you can't disconnect wallet. If we bring the header's stacking context higher, the modal close button will not work when scrolled and so on, and so on...
There are way too many stacking contexts which got clamped and I'm afraid it's impossible to pass it through without changing the DOM structure. I played around with various `z-index` values over elements and fixing one issue only brings another. 

The desired solution would be to render the Sidebar component alongside with the Modal - in the same portal node. It will cause the Sidebar to inherit the stacking context of the modal so it will be operable without the need of enforcing it manually with `z-index` property.

As a temporary solution I'd revert the change to `1450` keeping in mind now the Header is inoperable so it needs to be fixed as well. I'd suggest to leave a comment in code and handle this fix in separate PR. The priority of changing the account with open modal is lower than the ability of closing modal on smaller screens. You can change the account elsewhere, however with inoperable Modal's close button you get stuck.

https://github.com/user-attachments/assets/6436b4a9-21b2-4ab9-880f-eba57d62f06b

dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx
```diff
@@ -53,7 +53,10 @@ export default function DepositBTCModal() {
 
   const onDepositBTCError = useCallback(
     (error: unknown) => {
-      if (eip1193.didUserRejectRequest(error)) {
+      if (
+        eip1193.didUserRejectRequest(error) ||
+        ledgerLive.didUserRejectRequest(error)
```
nkuba, line 58: Can we combine these two with one common utils function?

dapp/src/components/TransactionModal/ActiveUnstakingStep/SignMessageModal.tsx
```diff
@@ -106,7 +111,10 @@ export default function SignMessageModal() {
     (error: unknown) => {
       if (!sessionIdToPromise[sessionId.current].shouldOpenErrorModal) return
 
-      if (eip1193.didUserRejectRequest(error)) {
+      if (
+        eip1193.didUserRejectRequest(error) ||
+        ledgerLive.didUserRejectRequest(error)
```
nkuba, line 116: Can we combine these two with one common utils function?

dapp/src/constants/wallets.ts
```diff
@@ -31,15 +31,24 @@ const XVERSE: WalletInfo = {
   },
 }
 
+const LEDGER: WalletInfo = {
```
nkuba, line 34:
```suggestion
const LEDGER_LIVE: WalletInfo = {
```

nkuba (commented):
I have a problem with the withdrawal flow. When I click the "Withdraw" button, the modal gets switched to Deposit form.

https://github.com/user-attachments/assets/19db3596-fb6e-4844-83da-94ebc8992c75

dapp/src/utils/orangekit/ledger-live/bitcoin-provider.ts
kkosiorowska: This error occurs when the user closes the transaction/sign message window see the video below. So I wonder if we can't add this error message to the `tryRequest` function to remove the `ledgerLive.didUserRejectRequest` function. Of course, this is not a blocker but some later improvement.

https://github.com/user-attachments/assets/4ade40a8-618b-4198-9d3c-d065af1a4869

dapp/manifests/ledger-live/ledger-live-manifest-mainnet.json
```diff
@@ -0,0 +1,31 @@
+{
+  "id": "acre-mainnet",
```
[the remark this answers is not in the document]
r-czajkowski: 49c77ea

dapp/src/store/wallet/walletSlice.ts
```diff
@@ -101,7 +103,7 @@ export const walletSlice = createSlice({
       const activityId = action.payload
       delete state.latestActivities[activityId]
     },
-    resetState: () => initialState,
+    resetState: (state) => ({ ...initialState, address: state.address }),
```
kkosiorowska, line 106: Why don't we also reset the address state?

dapp/src/theme/utils/zIndices.ts
nkuba: The temporary fix works. I added a ticket for the proper fix to the backlog https://github.com/thesis/acre/issues/832.

nkuba (approved):
It seems that there are no more blocking comments. This PR introduces many improvements to both Standalone and Ledger Live dApp and blocks other PRs.

I'll merge this PR, but @r-czajkowski please address all the unresolved comments in a follow-up PR and @kkosiorowska feel free to propose the improvements.

dapp/src/components/TransactionModal/ActiveUnstakingStep/SignMessageModal.tsx
r-czajkowski: 8b364a0

dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx
r-czajkowski: 8b364a0

PR #716, opened 2024-08-28
Solidity v1.1.0 release

nkuba:
This PR contains artifacts of contracts deployed on the mainnet in the v1.1.0 release. It also updates the version in package.json.

> [!IMPORTANT]  
> Please ignore failing integration tests, as these will be fixed in https://github.com/thesis/acre/pull/718

PR #717, opened 2024-08-28
MezoAllocator Withdrawals Improvements

nkuba:
Depends on: https://github.com/thesis/acre/pull/716

In this PR we introduce a couple of improvements to the MezoAllocator contract related to withdrawals from the Mezo Portal.

### Check if the withdrawal amount exceeds the deposit balance (1b95419737ea5d012647e0558e61a11b08feec86)

Here we improve the withdrawal function to check if the withdrawal amount exceeds the deposit balance.
If the withdrawal amount exceeds the deposit balance, the function will revert with a custom error message.
Before these changes, since the Portal contract changed the handling of partial and full withdrawals the withdrawal function reverted on `depositBalance -= uint96(amount);` with:
```
panic code 0x11 (Arithmetic operation overflowed outside of an unchecked block)
```

### Include unallocated MezoAllocator balance in `withdraw` and `releaseDeposit` (1b9ba17fe3df251445dc36b6b6452c3dd242ccb0, 0440580dcfcc879eb417d07033c30ed33af3e3a2)

In bb42bce291dc77a64d77a6500492ef23714aeb03 we included the current balance of the MezoAllocator contract in the `totalAssets` function result. The balance can come from a rewards transfer or any donation.

We haven't considered this balance in the `withdraw` function, which could lead to blocking the last user withdrawing the funds, as the balance of assets owned by them calculated based on `totalAssets` function, couldn't be fully withdraw due to unallocated balance being stuck in the `MezoAllocator` contract. In 1b9ba17fe3df251445dc36b6b6452c3dd242ccb0 we introduce possibility to withdraw funds from the unallocated contract balance.

For the `releaseDeposit` function we could face a problem when the Mezo Portal deposit is already released but some tokens were donated to the MezoAllocator contract, since the `Porta.withdraw` function would be reverting as there is no deposit, we won't be able to transfer the unallocated funds in the same call. We fixed it in 0440580dcfcc879eb417d07033c30ed33af3e3a2.

> [!IMPORTANT]  
> Please ignore failing integration tests, as these will be fixed in https://github.com/thesis/acre/pull/718

### DIFF
--- a/solidity/contracts/MezoAllocator.sol
+++ b/solidity/contracts/MezoAllocator.sol
@@ -231,25 +231,32 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
     function withdraw(uint256 amount) external {
         if (msg.sender != address(stbtc)) revert CallerNotStbtc();
 
-        emit DepositWithdrawn(depositId, amount);
+        uint256 unallocatedBalance = tbtc.balanceOf(address(this));
 
-        if (amount < depositBalance) {
-            mezoPortal.withdrawPartially(
-                address(tbtc),
-                depositId,
-                uint96(amount)
-            );
-        } else if (amount > depositBalance) {
-            revert WithdrawalAmountExceedsDepositBalance(
-                amount,
-                depositBalance
-            );
-        } else {
-            mezoPortal.withdraw(address(tbtc), depositId);
+        if (amount > unallocatedBalance) {
+            uint256 amountToWithdraw = amount - unallocatedBalance;
+
+            emit DepositWithdrawn(depositId, amountToWithdraw);
+
+            if (amountToWithdraw < depositBalance) {
+                mezoPortal.withdrawPartially(
+                    address(tbtc),
+                    depositId,
+                    uint96(amountToWithdraw)
+                );
+            } else if (amountToWithdraw > depositBalance) {
+                revert WithdrawalAmountExceedsDepositBalance(
+                    amountToWithdraw,
+                    depositBalance
+                );
+            } else {
+                mezoPortal.withdraw(address(tbtc), depositId);
+            }
+
+            // slither-disable-next-line reentrancy-no-eth
+            depositBalance -= uint96(amountToWithdraw);
         }
 
-        // slither-disable-next-line reentrancy-no-eth
-        depositBalance -= uint96(amount);
         tbtc.safeTransfer(address(stbtc), amount);
     }
 
--- a/solidity/contracts/MezoAllocator.sol
+++ b/solidity/contracts/MezoAllocator.sol
@@ -252,13 +252,13 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
                     depositId,
                     uint96(amountToWithdraw)
                 );
-            } else if (amountToWithdraw > depositBalance) {
+            } else if (amountToWithdraw == depositBalance) {
+                mezoPortal.withdraw(address(tbtc), depositId);
+            } else {
                 revert WithdrawalAmountExceedsDepositBalance(
                     amountToWithdraw,
                     depositBalance
                 );
-            } else {
-                mezoPortal.withdraw(address(tbtc), depositId);
             }
 
             // slither-disable-next-line reentrancy-no-eth
--- a/solidity/contracts/MezoAllocator.sol
+++ b/solidity/contracts/MezoAllocator.sol
@@ -252,6 +252,7 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
                     depositId,
                     uint96(amountToWithdraw)
                 );
+                // slither-disable-next-line incorrect-equality
             } else if (amountToWithdraw == depositBalance) {
                 mezoPortal.withdraw(address(tbtc), depositId);
             } else {
--- a/solidity/contracts/MezoAllocator.sol
+++ b/solidity/contracts/MezoAllocator.sol
@@ -142,6 +142,12 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
     error MaintainerNotRegistered();
     /// @notice Reverts if the maintainer has been already registered.
     error MaintainerAlreadyRegistered();
+    /// @notice Reverts if the requested amount to withdraw exceeds the amount
+    ///         deposited in the Mezo Portal.
+    error WithdrawalAmountExceedsDepositBalance(
+        uint256 requestedAmount,
+        uint256 depositAmount
+    );
 
     modifier onlyMaintainer() {
         if (!isMaintainer[msg.sender]) {
@@ -232,6 +238,11 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
                 address(tbtc),
                 depositId,
                 uint96(amount)
+            );
+        } else if (amount > depositBalance) {
+            revert WithdrawalAmountExceedsDepositBalance(
+                amount,
+                depositBalance
             );
         } else {
             mezoPortal.withdraw(address(tbtc), depositId);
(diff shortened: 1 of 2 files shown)


solidity/test/MezoAllocator.test.ts
```diff
@@ -542,32 +733,142 @@ describe("MezoAllocator", () => {
...
+          before(async () => {
+            await tbtc.mint(await mezoAllocator.getAddress(), donationAmount)
+
+            tx = await mezoAllocator.connect(governance).releaseDeposit()
+          })
+
+          it("should emit DepositReleased event", async () => {
+            await expect(tx)
+              .to.emit(mezoAllocator, "DepositReleased")
+              .withArgs(1, depositAmount)
```
r-czajkowski, line 851: Can we extract `1` to variable`const depositId = 1`? At first sight, it's hard to understand what `1` means (I had to check the `DepositReleased` event).

solidity/test/MezoAllocator.test.ts
```diff
@@ -542,32 +733,142 @@ describe("MezoAllocator", () => {
...
+              tbtc,
+              [mezoPortal, mezoAllocator, stbtc],
+              [-depositAmount, -donationAmount, depositAmount + donationAmount],
+            )
+          })
+
+          it("should call MezoPortal.withdraw function", async () => {
+            await expect(tx)
+              .to.emit(mezoPortal, "WithdrawFully")
+              .withArgs(await tbtc.getAddress(), 1)
```
r-czajkowski, line 870: Same as above.

solidity/test/MezoAllocator.test.ts
```diff
@@ -542,32 +733,142 @@ describe("MezoAllocator", () => {
...
+          let tx: ContractTransactionResponse
+
+          before(async () => {
+            tx = await mezoAllocator.connect(governance).releaseDeposit()
+          })
+
+          it("should emit DepositReleased event", async () => {
+            await expect(tx)
+              .to.emit(mezoAllocator, "DepositReleased")
+              .withArgs(1, depositAmount)
```
r-czajkowski, line 812: I noticed that we do not define the deposit id as a variable in most places. In my opinion, extracting the value to a variable with a good name will improve readability but ofc non-blocking.

solidity/contracts/MezoAllocator.sol
```diff
@@ -225,20 +231,32 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
...
-            );
-        } else {
-            mezoPortal.withdraw(address(tbtc), depositId);
+            // slither-disable-next-line reentrancy-no-eth
+            depositBalance -= uint96(amountToWithdraw);
         }
 
-        // slither-disable-next-line reentrancy-no-eth
-        depositBalance -= uint96(amount);
         tbtc.safeTransfer(address(stbtc), amount);
```
pdyraga, line 260: Shouldn't we emit some event informing about withdrawing from the allocator? `DepositWithdrawn` is only about withdrawing from the Portal.

solidity/contracts/MezoAllocator.sol
```diff
@@ -225,20 +231,32 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
...
+
+            emit DepositWithdrawn(depositId, amountToWithdraw);
+
+            if (amountToWithdraw < depositBalance) {
+                mezoPortal.withdrawPartially(
+                    address(tbtc),
+                    depositId,
+                    uint96(amountToWithdraw)
+                );
+            } else if (amountToWithdraw > depositBalance) {
```
pdyraga, line 247: Not a blocker and nothing wrong, just a heads up for future changes. In `Portal` there is a way someone can [donate](https://github.com/thesis/mezo-portal/blob/f563927ee6c5b670599bd7d0973194b5537b3c4f/solidity/contracts/Portal.sol#L669) your deposit. In this case, it may happen `depositBalance` tracked separately will be lower than what's actually in the Portal.

pdyraga (commented):
I have not reviewed the tests, just Solidity changes. They look solid to me, left two non-blocking comments/questions.

solidity/contracts/MezoAllocator.sol
```diff
@@ -225,20 +231,32 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
...
+                revert WithdrawalAmountExceedsDepositBalance(
+                    amountToWithdraw,
+                    depositBalance
+                );
+            } else {
+                mezoPortal.withdraw(address(tbtc), depositId);
```
dimpar, line 253: Scenario:
`unallocatedBalance` = 0
`amount` = 10tBTC
`amountToWithdraw` == `depositBalance`
allocator calls `mezoPortal.withdraw` but there's a [fee in Mezo Portal](https://github.com/thesis/mezo-portal/blob/main/solidity/contracts/Portal.sol#L544). Say it's 0.1. So `MezoPortal` will return 9.9tBTC
Now, the allocator get's 9.9 which is less than 10. `tbtc.safeTransfer(address(stbtc), amount)` will revert in this case. Am I missing something? Is that known?

solidity/contracts/MezoAllocator.sol
```diff
@@ -225,20 +231,32 @@ contract MezoAllocator is IDispatcher, Ownable2StepUpgradeable {
...
 
-        emit DepositWithdrawn(depositId, amount);
+        uint256 unallocatedBalance = tbtc.balanceOf(address(this));
+
+        if (amount > unallocatedBalance) {
+            uint256 amountToWithdraw = amount - unallocatedBalance;
+
+            emit DepositWithdrawn(depositId, amountToWithdraw);
+
+            if (amountToWithdraw < depositBalance) {
```
dimpar, line 241: How about changing the order of the logic?
```
 if (amountToWithdraw < depositBalance) {
            mezoPortal.withdrawPartially(
                address(tbtc),
                depositId,
                uint96(amountToWithdraw)
            );
        } else if (amountToWithdraw == depositBalance) {
            mezoPortal.withdraw(address(tbtc), depositId);
        } else {
            revert WithdrawalAmountExceedsDepositBalance(
                amountToWithdraw,
                depositBalance
            );
        }
```

```
if(..) {
  valid case
} else if (..) {
  valid case
} else {
  error
}
```

instead of error being in the middle. Might be somewhat cleaner. Up to you, not a blocker.

solidity/test/MezoAllocator.test.ts
nkuba: 93a781f0aced7c7a5f2a7c7e0cb3be0a0bf47077

solidity/test/MezoAllocator.test.ts
nkuba: 93a781f0aced7c7a5f2a7c7e0cb3be0a0bf47077

solidity/contracts/MezoAllocator.sol
nkuba: dcbe752a3593b3d4af352a61e76926bb0dd0d3e6

solidity/contracts/MezoAllocator.sol
nkuba: The fee in MezoPortal is charged only for deposits with a minted receipt. For deposits made by the MezoAllocator, the withdrawal balance is expected to always match the deposit balance.
@pdyraga are there any plans to further change Mezo Portal withdrawals that would affect the MezoAllocator contract?

solidity/contracts/MezoAllocator.sol
nkuba: 7b97aac7f45a59b73a9e73f505e3d9e1aa1f922a

solidity/contracts/MezoAllocator.sol
pdyraga: There are no other fees and no other plans for fees in the Portal than the fee when someone minted the receipt token like stBTC. In this case, the fee is paid during the withdrawal. There are also fees from tBTC bridge like the redemption fee and minting fee but they do not affect Acre's MezoAllocator to my knowledge.

dimpar (approved):

r-czajkowski (approved):

r-czajkowski (approved):

Issue #719, opened 2024-09-02
dApp Redesign


nkuba:
Figma: [link](https://www.figma.com/design/j2MModNsoiOYNuBBXbWiM8/Web-dApp?node-id=16848-27257&node-type=FRAME&m=dev)

```[tasklist]
### Tasks
- [ ] Change the navigation menu in the header
- [ ] Remove the Season landing page
- [ ] Change the "Connect wallet" button background
- [ ] Add Tooltip to the beehive component
- [ ] Rename Beehive to L2 Rewards
- [ ] Change behavior for the not connected user (display empty state)
```

PR #720, opened 2024-09-03
Redesign dapp

kkosiorowska:
Closes: https://github.com/thesis/acre/issues/719

**Tasks:**

- [x] Change the navigation menu in the header 
- [x] Remove the Season landing page
- [x] Change the "Connect wallet" button background
- [x] Add Tooltip to the beehive component
- [x] Rename Beehive to L2 Rewards
- [x] Change behavior for the not connected user (display empty state)


dapp/src/components/Header/Navigation/Navigation.tsx
```diff
@@ -1,20 +1,29 @@
...
+function Navigation(props: BoxProps) {
   return (
-    <Box as="nav" {...restProps}>
+    <Box as="nav" {...props}>
       <HStack as={List} spacing={5} ml={12}>
-        {items.map((item) => (
-          <NavigationItem key={item.href} {...item} />
+        {NAVIGATION_ITEMS.map((item) => (
+          <NavigationItem
+            key={isString(item.to) ? item.to : item.to.pathname}
```
kpyszkowski, line 24: The type definition of `NAVIGATION_ITEMS` variable makes this line unnecessarily complicated.
The `to` type is always a `string` so the type check is redundant as you remove/simplify the type definition.

The given solution is much more readable in my opinion. The type definition is more specific and the need of programmatic type check disappears. The `label` is optional, you can keep `children` if you want but the `label` is more intuitive.

```tsx
type NavigationItemType = Omit<NavigationItemProps, 'children' | 'to'> {
  label: string
  to: string
}

const NAVIGATION_ITEMS: NavigationItemType[] = [
// ...
]

// ...

{NAVIGATION_ITEMS.map(({label, to, ...itemProps}) => (
  <NavigationItem  key={to} {...itemProps}>
    {label}
  </NavigationItem>
))}
```
 What do you think?

dapp/src/pages/DashboardPage/UsefulLinks.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+import { VStack } from "@chakra-ui/react"
+import ButtonLink from "#/components/shared/ButtonLink"
+import { EXTERNAL_HREF } from "#/constants"
+
+export default function UsefulLinks() {
+  return (
+    <VStack>
+      {[
+        { label: "Documentation", url: EXTERNAL_HREF.DOCS },
```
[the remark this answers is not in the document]
kkosiorowska: In this case, we are using a simple button which is a link so let's use an `href` here to not mix naming.

https://github.com/thesis/acre/pull/720/commits/3d69111631c8c5d91f4b73ad3e4e8c54bf1bbfe6

dapp/src/components/Header/Navigation/Navigation.tsx
kkosiorowska: > The type definition of NAVIGATION_ITEMS variable makes this line unnecessarily complicated.
The to type is always a string so the type check is redundant as you remove/simplify the type definition.

I would not limit the type for navigation items and let it be simply of type `NavigationItemProps`. This approach seems to me more intuitive and consistent with the rest instead of creating another slightly modified type that is used only in one place. 

I simplified the code a bit to make it more clear. - https://github.com/thesis/acre/pull/720/commits/a4c7ebeb8aea1bf96c7471dca9b8001c372f9087


> The given solution is much more readable in my opinion. The type definition is more specific and the need of programmatic type check disappears. The label is optional, you can keep children if you want but the label is more intuitive.

Declaring the `label` as a `string` limits later editing of navigation elements. If we want to add an icon to the navigation element, we will do it easily using `children`. Therefore, let's stay with the use of `children`.

kpyszkowski (approved):

PR #721, opened 2024-09-04
 Center the dApp header

kkosiorowska:
**Before**
image

**After**
Screenshot 2024-09-04 at 11 14 18

### DIFF
--- a/dapp/src/theme/utils/semanticTokens.ts
+++ b/dapp/src/theme/utils/semanticTokens.ts
@@ -3,9 +3,10 @@ export const semanticTokens = {
     header_height: 24,
     modal_shift: "9.75rem", // 156px
     dashboard_card_padding: 5,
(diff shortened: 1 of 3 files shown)


dapp/src/theme/utils/semanticTokens.ts
```diff
@@ -3,9 +3,10 @@ export const semanticTokens = {
     header_height: 24,
     modal_shift: "9.75rem", // 156px
     dashboard_card_padding: 5,
+    page_content_padding_x: 10,
```
kpyszkowski, line 6: Non blocking
I have mixed feelings about these "tokens". We don't even reuse it elsewhere.
It only makes the need to jump between files as I'd like to investigate the value read from component JSX.

kpyszkowski (approved):
One non blocking.
Perhaps future improvement if you agree.

PR #723, opened 2024-09-06
Improvements to the mintDebt and repayDebt functions in stBTC contract

nkuba:
### Fix returns in mintDebt and repayDebt functions (4f67f035076dd6dd37461d14a0fd771bbf9645f1)

The `mintDebt` and `repayDebt` functions are expected to return the amount of assets calculated based on the provided amount of shares:
```solidity
) public whenNotPaused returns (uint256 assets) {
```
but they were returning the amount of shares instead:
```solidity
return shares;
```
We fix this problem and add tests to cover these cases.

### Fix rounding of shares to assets conversions in debt minting (490b1790154e8c7842541f08ea9b1a11097e2437)

Previously, the shares-to-assets conversion was rounding down, which could have resulted in the minted debt being against the vault's interest.

Suppose a scenario where the vault has a 10% loss, meaning that each share is worth 0.9 assets.
If a user mints 1 share, the debt registered due to rounding down would be 0.
This can be used in a griefing attack, where a user could mint shares without registering debt.

To fix the issue, we now round up the shares to assets conversion in the `mintDebt` function.

Additionally, for the `repayDebt` function, we keep rounding down the shares to assets to avoid the opposite issue, where a user could repay more debt than they should.

It may happen that conversion in `repayDebt` returns `0`, which should be validated by the caller. Since the interface used by the Mezo Portal contract `burnReceipt` function doesn't consider the return value of `repayDebt`, we are performing the validation in the `burnReceipt` function. We added a similar validation in the `mintReceipt` function for consistency.

### DIFF
--- a/solidity/contracts/stBTC.sol
+++ b/solidity/contracts/stBTC.sol
@@ -326,8 +326,6 @@ contract stBTC is ERC4626Fees, PausableOwnable {
 
         // Mint the shares to the receiver.
         super._mint(receiver, shares);
-
-        return shares;
     }
 
     /// @dev This function proxies `mintDebt` call and provides compatibility
@@ -368,8 +366,6 @@ contract stBTC is ERC4626Fees, PausableOwnable {
 
         // Burn the shares from the debtor.
         super._burn(msg.sender, shares);
-
-        return shares;
     }
 
     /// @notice This function proxies `repayDebt` call and provides
--- a/solidity/contracts/stBTC.sol
+++ b/solidity/contracts/stBTC.sol
@@ -125,6 +125,10 @@ contract stBTC is ERC4626Fees, PausableOwnable {
     /// Reverts if the dispatcher address is the same.
     error SameDispatcher();
 
+    /// Reverts if a conversion between shares and assets results in zero, which
+    /// may happen for small amounts division with rounding.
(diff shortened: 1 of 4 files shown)


solidity/contracts/stBTC.sol
```diff
@@ -326,17 +330,18 @@ contract stBTC is ERC4626Fees, PausableOwnable {
...
     }
 
     /// @dev This function proxies `mintDebt` call and provides compatibility
     ///      with Mezo IReceiptToken interface.
     function mintReceipt(address to, uint256 amount) external {
-        mintDebt(amount, to);
+        uint256 assets = mintDebt(amount, to);
+        if (assets == 0) {
+            revert ConvertedToZero(amount, assets);
+        }
```
jagodarybacka, line 341: So do I understand it correctly that in `mintReceipt` the `ConvertedToZero` error will be triggered when user will request too low amount of stBTC in the context of a given state of contract, right?

And alike - in `burnReceipt` user can trigger `ConvertedToZero` by requesting to burn a low amount of asset.

But what if that's all that was minted by this user? They will not be able to burn their stBTC?

nkuba:
I'm closing this PR as the debt minting feature will be removed soon.

PR #725, opened 2024-09-12
Disable the Deposit button for Xverse users on production

kkosiorowska:
Currently, there are some problems with the execution of the deposit for the Xverse wallet. So let's disable deposit flow for Xverse users on production.

### Testing

- [x] The deposit button should be disabled when the user is connected by Xverse wallet.
- [x] When the user is not connected and clicks the deposit button, the user can connect, but the deposit window does not open when the user selects the Xverse wallet.
- [x] The changes do not affect behavior for other wallets.
- [x] The tooltip with information appears over the disabled button as the user moves cursor over


kpyszkowski (approved):
LGTM

nkuba (approved):

Issue #51, opened 2023-11-27
 Account context caching


kkosiorowska:
For some applications like Lido in Ledger Live after refreshing the app, the user is still connected with the dApp. Make sure our dApp works the same way.

nkuba:
Solved in https://github.com/thesis/acre/pull/708/commits/c4e698f3376d661ff569b86ace36f272ab683957

PR #742, opened 2024-09-27
 Init "0-click connection" flow for the Ledger Live App

kkosiorowska:
Closes https://github.com/thesis/acre/issues/51

### What has been done:

- Trigger the connect wallet flow when application starts 
- Cache the last used account in local storage


dapp/src/hooks/useLastUsedBtcAddress.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback } from "react"
+import useLocalStorage from "./useLocalStorage"
+
+export const LAST_USED_BTC_ADDRESS_KEY = "lastUsedBtcAddress"
```
nkuba, line 4: Can we keep the `LAST_USED_BTC_ADDRESS_KEY` unexported? We could define helper functions to read and write the last used btc address in the storage. Then we could use the read function in `dapp/src/wagmiConfig.ts`.

nkuba (approved):
We will follow-up in the feature branch.

dapp/src/hooks/useLastUsedBtcAddress.ts
nkuba: Moved the discussion to https://github.com/thesis/acre/pull/708#discussion_r1784027460

Issue #726, opened 2024-09-16
Improve interaction with wallets


kkosiorowska:
Recently, several issues with wallet interaction have been noted. These are a few issues that should be fixed:

### Unisat

- [ ] After disconnecting the account, the wallet window opens. But the status of the steps is closed.

https://github.com/user-attachments/assets/660646ae-8404-42ea-b67d-9f545ceaf41e

**Expected result:** The window should not open after disconnection.

### All walets

- [x] When the wallet connection action is rejected, a new window opens. But the status of the steps is not shown.

https://github.com/user-attachments/assets/394e45c3-64f8-476d-adc6-84c8f0c23dee

**Expected result:** We probably shouldn't trigger a new connection action. To be confirmed with @SorinQ

PR #744, opened 2024-10-01
Bug fixes for Unisat

kkosiorowska:
Closes https://github.com/thesis/acre/issues/726
Closes https://github.com/thesis/acre/issues/751

This PR fixes the issue of the Unisat windows showing up after disconnection. The problem started to occur from v1.4.5 Unisat version. This is related to the implementation of the [disconnect](https://github.com/hemilabs/ui-monorepo/pull/473/files) method. I'm not sure what exactly is causing the problems but it seems that when the user is disconnected and the `onDisconnect` function is called once again the wallet window opens.

Two issues were identified:

- The `useDisconnectWallet` hook was run with outdated states by this `onDisconnect` was run once again. 

**Solution:** Let's make sure that the user's data are reset after disconnection. Run the reset function in the `onSuccess` callback. 
-  The `handleAccountsChanged` function runs even when a user disconnects from the dApp level. Then when the user is disconnected and the `onDisconnect` is called again unisat will open a wallet window. 

**Solution:** Let's disable the disconnect feature from the wallet level for now. Hook `useAccountsChangedUnisat` should be moved to OrangeKit anyway and fixed at that point. The solution isn't perfect, but it improves the Unisat wallet experience a lot.

### UI

**Before**

https://github.com/user-attachments/assets/3d0f4b7d-aa2f-4da6-9953-7181e30577d6

https://github.com/user-attachments/assets/f253eb6c-d243-43eb-8f63-cfe4bb196194

**After**

https://github.com/user-attachments/assets/3042c7b4-ec45-4125-ad53-e4185b549c54

### Testing

- [x] Accept dapp connection, Cancel SIWE, Close modal window
- [x] Logout
- [x] Change wallet addresses in the extension


r-czajkowski (approved):

michalinacienciala:
Tested the cases mentioned in the description, they behaved ok.

I also tested such case:
1. Fully log in to Acre with Unisat account
2. Disconnect from Acre dapp in Unisat
3. Wait until dapp refreshes the data or open deposit form

Result:
Unisat window asking for connection to Acre appears. When user chooses to cancel, a new window gets opened (this continues until user cancels for the 4th time or accepts the connection). Ideally I think we should log out the user from the portal when they choose to disconnect the dapp within the wallet extension. Not sure if this is a scope of this PR or if it should be handled separately.

kkosiorowska:
> Tested the cases mentioned in the description, they behaved ok.
> 
> I also tested such case:
> 
> 1. Fully log in to Acre with Unisat account
> 2. Disconnect from Acre dapp in Unisat
> 3. Wait until dapp refreshes the data or open deposit form
> 
> Result: Unisat window asking for connection to Acre appears. When user chooses to cancel, a new window gets opened (this continues until user cancels for the 4th time or accepts the connection). Ideally I think we should log out the user from the portal when they choose to disconnect the dapp within the wallet extension. Not sure if this is a scope of this PR or if it should be handled separately.

This PR disables event listening when the user decides to disconnect from the wallet level. This is the simplest solution to fix the problem of multiple Unisat windows. The solution is not perfect but makes the user experience much better.  This part should be fixed when event listeners are moved to OrangeKit. Created issue - https://github.com/thesis/orangekit/issues/132

michalinacienciala (approved):

PR #745, opened 2024-10-01
Update text from 'Deposit more' to 'Deposit' for new users

kkosiorowska:
https://github.com/user-attachments/assets/6120eba2-3233-441f-b433-31575e215aaa


kpyszkowski (approved):
LGTM

Issue #753, opened 2024-10-02
Add switch account button in Ledger Live dApp


nkuba:
The user should be able to switch accounts with one button in Ledger Live. Currently, they have to press "Disconnect" and "Connect account"

PR #760, opened 2024-10-09
Add switch account button

r-czajkowski:
Closes: #753 

This PR adds a switch account button according to the Figma designs. Users must sign a message to create a session when they want to change an account.

obraz

https://github.com/user-attachments/assets/8c72dc1c-9ef9-4c4c-9ca5-114a6b23d847

### DIFF
--- a/dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
+++ b/dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
@@ -63,7 +63,6 @@ export default function ConnectWalletButton({
     onConnect,
     onDisconnect,
     status: connectionStatus,
-    reconnectStatus,
   } = useWallet()
   const { signMessageStatus, resetMessageStatus, signMessageAndCreateSession } =
     useSignMessageAndCreateSession()
@@ -250,11 +249,7 @@ export default function ConnectWalletButton({
                   Requires 2 actions:
                 </TextMd>
                 <ConnectWalletStatusLabel
-                  status={
-                    connectionStatus === "idle" && reconnectStatus !== "idle"
-                      ? reconnectStatus
-                      : connectionStatus
-                  }
+                  status={connectionStatus}
                   label={`Connect ${isEmbed ? "account" : "wallet"}`}
                 />
                 <ConnectWalletStatusLabel
--- a/dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
+++ b/dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
@@ -63,6 +63,7 @@ export default function ConnectWalletButton({
     onConnect,
     onDisconnect,
     status: connectionStatus,
+    reconnectStatus,
   } = useWallet()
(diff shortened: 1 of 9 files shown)


kkosiorowska (commented):
When I select the same account the data is loading so long. I wonder if we can somehow improve it because data should be already fetched. 

Actually, I think that the user is stuck on this view because in my case the data after a long time did not load.

Screenshot 2024-10-11 at 09 53 02

kkosiorowska (commented):
During the tests, I noticed two things:
1. Step one (“Connect account”) does not show in progress status when I want to change the account.
2. When I reject the connection and want to select the account again, I can't do it.

https://github.com/user-attachments/assets/a35cf5c2-7c51-4e76-9ac4-03ad646353a9

r-czajkowski:
> During the tests, I noticed two things:
> 
>     1. Step one (“Connect account”) does not show in progress status when I want to change the account.
> 
>     2. When I reject the connection and want to select the account again, I can't do it.
> 
> 
> Screen.Recording.2024-10-16.at.12.47.09.mov

Should be good now. See 71c79de 4e9541b.

dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
```diff
@@ -54,13 +55,15 @@ export default function ConnectWalletButton({
...
   onSuccess,
+  isReconnecting,
 }: ConnectWalletButtonProps) {
   const { isEmbed } = useIsEmbed()
   const {
     address,
     onConnect,
     onDisconnect,
     status: connectionStatus,
+    reconnectStatus,
```
kkosiorowska, line 66: Non-blocking comment

The `onConnect` function works in normal mode and in reconnecting mode. So I think we don't need to return two independent statuses but just one.

dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
```diff
@@ -239,7 +250,11 @@ export default function ConnectWalletButton({
                   Requires 2 actions:
                 </TextMd>
                 <ConnectWalletStatusLabel
-                  status={connectionStatus}
+                  status={
+                    connectionStatus === "idle" && reconnectStatus !== "idle"
+                      ? reconnectStatus
+                      : connectionStatus
```
kkosiorowska, line 256: As I mentioned above we could have one status and include this logic in `useWallet`.

kkosiorowska (approved):
LGTM, I only left non-blocking comments.

dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
r-czajkowski: aecbe9a

dapp/src/components/ConnectWalletModal/ConnectWalletButton.tsx
r-czajkowski: aecbe9a

Issue #759, opened 2024-10-07
Set referral for Ledger Live


nkuba:
> The `embed` parameter is not boolean. We will use it to pass the partner integration ID what will have to map to a referral ID. For Ledger Live we will have `embed=ledger-live` and we need to map it to the referral ID assigned for Ledger Live, which is number `2083` (source: [Coda page](https://coda.io/d/Acre_d3Vs1FswTQA/Referrals_sudl9wjh#_lubtctD5))

_Originally posted by @nkuba in https://github.com/thesis/acre/pull/708#discussion_r1775023581_

PR #761, opened 2024-10-09
Set referral code for embedded app

r-czajkowski:
Depends on: #760
Closes: #759 

This PR sets the referral code based on the embedded app.


dapp/src/hooks/useReferral.ts
```diff
@@ -14,39 +12,19 @@ type UseReferralReturn = {
...
-    if (referralProgram.isValidReferral(convertedReferral)) {
-      setReferral(convertedReferral)
-    } else {
-      console.error("Incorrect referral")
-      setReferral(null)
-      openModal(MODAL_TYPES.UNEXPECTED_ERROR, {
-        withCloseButton: false,
-        closeOnEsc: false,
-      })
-    }
```
kkosiorowska, line 48: If you remove this part, the referral for the standalone dApp is not set correctly. Are we sure we want to disable the referral program for the standalone page?

dapp/src/components/Header/ConnectWallet.tsx
kkosiorowska, line 1: It seems that changes for this file should be in https://github.com/thesis/acre/pull/760

nkuba:
Does this PR have to use [`add-change-acc-btn`](https://github.com/thesis/acre/tree/add-change-acc-btn) as the base branch? Can't it go directly to the [`ledger-live-updates`](https://github.com/thesis/acre/tree/ledger-live-updates) branch?

dapp/src/components/Header/ConnectWallet.tsx
r-czajkowski: I use `referralProgram.isEmbedApp(embeddedApp)` in `ConnectWallet` which I added in this PR. I can move the conditional rendering of the change account button to #760 and update the condition in this PR by using `referralProgram.isEmbedApp(embeddedApp)`.

r-czajkowski:
> Does this PR have to use [`add-change-acc-btn`](https://github.com/thesis/acre/tree/add-change-acc-btn) as the base branch? Can't it go directly to the [`ledger-live-updates`](https://github.com/thesis/acre/tree/ledger-live-updates) branch?

Here we hide the change account button for the embedded app and use some of the functions I added in this PR.

dapp/src/components/Header/ConnectWallet.tsx
kkosiorowska: Ah, okay. Let's leave it as it is.

dapp/src/components/Header/ConnectWallet.tsx
```diff
@@ -111,22 +115,24 @@ export default function ConnectWallet() {
...
-            <IconButton
-              variant="ghost"
-              aria-label="Change account"
-              icon={<Icon as={IconUserCode} boxSize={5} />}
-              px={2}
-              boxSize={5}
-              onClick={() => handleConnectWallet(true)}
-            />
-          </Tooltip>
+          {isChangeAccountFeatureSupported(embeddedApp) && (
```
kkosiorowska, line 118: Non-blocking comment

I think we don't need to create a separate function. We can simply use `isEmbed`.

kkosiorowska (approved):
I tested the solution and it seems to work well 🚀

dapp/src/components/Header/ConnectWallet.tsx
r-czajkowski: I think it depends on the embedded app. For example, we can support it in Ledger Live but not in Xverse.

PR #762, opened 2024-10-09
Multi Asset Vault contract implementation

nkuba:
In this PR, we implement the MultiAssetVault contract. This contract manages deposits and withdrawals of multiple assets. It is designed to be used in addition to the stBTC vault to allow users to deposit assets other than tBTC.

The stBTC token will not be generated for the assets deposited in the MultiAssetVault.

Deposits created in this vault are routed directly to the Mezo Portal, which generates a unique identifier for each deposit.

The deposit can be withdrawn by providing the deposit ID. The deposit is fully withdrawn from the Mezo Portal, and the assets are transferred to the user.

Partial withdrawals are not supported. The user must withdraw the entire deposit.


solidity/contracts/MultiAssetVault.sol
```diff
@@ -0,0 +1,240 @@
...
+        uint256 initialBalance = IERC20(asset).balanceOf(address(this));
+
+        mezoPortal.withdraw(address(asset), depositId);
+
+        uint256 withdrawnAmount = IERC20(asset).balanceOf(address(this)) -
+            initialBalance;
+
+        if (withdrawnAmount != depositedAmount) {
+            revert UnexpectedWithdrawnAmount(withdrawnAmount, depositedAmount);
+        }
```
pdyraga, line 224: Just a heads up this is currently possible only in two scenarios that are not triggerable for `MultiAssetVault` contract: 
- WBTC -> tBTC migration in the Mezo Portal when decimals are switched.
- Deposit receipt (stBTC) minted and repaid for non-zero fees.

solidity/deploy/42_register_solvbtc.ts
```diff
@@ -0,0 +1,37 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types"
```
pdyraga, line 1: Wouldn't it be easier to accept the initial supported tokens in the `initialize` function, just [like in the Mezo Portal](https://github.com/thesis/mezo-portal/blob/f09b08ea26274a8c3694aabff82821f35f38dc16/solidity/contracts/Portal.sol#L478)? We could event hardcode them in the contract. We would reduce the deployment to just one script without the need to involve multisig to accept the ownership. 

`0x790Dda4c56b3c45d0e4514eDbAaBa30D7129c857` (the [`governance`](https://github.com/thesis/acre/blob/cb5f67c801b4263c3463b3b68352f11f47b8ae48/solidity/hardhat.config.ts#L101)) kept executing transactions in the past so we know it works and do not have to test it by accepting the ownership.

pdyraga (commented):
Left some early thoughts, I will want to take another look once the tests are ready.

solidity/contracts/MultiAssetVault.sol
```diff
@@ -0,0 +1,240 @@
...
+
+import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {IMezoPortal} from "./interfaces/IMezoPortal.sol";
+import {ZeroAddress} from "./utils/Errors.sol";
+
+/// @notice MultiAssetVault is a contract that allows users to deposit and withdraw
+///         multiple assets. It uses MezoPortal to deposit and withdraw assets.
+contract MultiAssetVault is Ownable2StepUpgradeable {
```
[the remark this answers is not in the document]
nkuba: be5138fee6aa0c8520fe33d8b88a3175b0504955

solidity/contracts/MultiAssetVault.sol
```diff
@@ -0,0 +1,240 @@
...
+        uint256 depositId = mezoPortal.depositCount();
+
+        // slither-disable-next-line reentrancy-events
+        emit DepositCreated(depositOwner, asset, depositId, amount);
+
+        // Register the deposit info.
+        // slither-disable-next-line reentrancy-benign
+        deposits[depositOwner][asset][depositId] = DepositInfo({
+            balance: amount
+        });
```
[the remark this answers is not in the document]
nkuba: ea75ab93febb9a2b6768ed922c0c97fbaf3344b2

solidity/contracts/MultiAssetVault.sol
nkuba: Yes, I'm aware that is not possible currently. I just wanted to protect us from unexpected upgrades or a bug in the Mezo Portal. With this check, we won't delete the deposit if something odd happens.

solidity/deploy/42_register_solvbtc.ts
nkuba: 4b7d865fbc5dfc254f51d91200cbc77f867d7998

solidity/contracts/MultiAssetVault.sol
pdyraga: Yep, this check makes sense. I only wanted to make sure we are in sync about the scenarios possible to trigger it.

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// @param asset Address of the asset to be removed from the supported assets.
+    function removeSupportedAsset(address asset) external onlyOwner {
+        if (!supportedAssets[asset]) {
+            revert AssetNotSupported();
+        }
+
+        emit SupportedAssetRemoved(asset);
+
+        supportedAssets[asset] = false;
+    }
```
jagodarybacka, line 148: We decided to not allow removing assets in Mezo to avoid complicating the business logic - do you expect to use this function? Why would you add and then remove the asset? What about the users that already deposited that asset?

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,278 @@
...
+        emit DepositCreated(depositOwner, asset, depositId, amount);
+
+        // Transfer the asset from the deposit owner to the vault and approve
+        // Mezo Portal to pull them.
+        IERC20(asset).safeTransferFrom(msg.sender, address(this), amount);
+        IERC20(asset).forceApprove(address(mezoPortal), amount);
+
+        // Create a deposit in Mezo Portal.
+        // 0 denotes no lock period for this deposit.
+        mezoPortal.deposit(address(asset), uint96(amount), 0);
```
jagodarybacka, line 191: What is that asset is not supported in Mezo? we just trust that this asset is supported?

jagodarybacka (commented):
Initial thoughts, nice to see Acre getting even closer to Mezo 🤗 

Note that you'll need to use our Supported Tokens mocks on Sepolia so we will need to share balances of these tokens between testing accounts but I guess that's fine, we will figure it out.

solidity/contracts/AcreMultiAssetVault.sol
nkuba: If the asset is removed new deposits can't be created for the given asset.
But existing depositors are able to withdraw, since we're not checking if the asset is supported in the withdraw function.

I was curious why Mezo Portal's withdraw function checks if the asset is supported?

solidity/contracts/AcreMultiAssetVault.sol
nkuba: If the asset is not supported in Mezo the transaction will revert with an error originated in the Mezo Portal contract.

solidity/deploy/41_deploy_multi_asset_vault.ts
```diff
@@ -0,0 +1,46 @@
...
+const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
+  const { deployments, helpers } = hre
+  const { deployer, governance } = await helpers.signers.getNamedSigners()
+
+  const mezoPortal = await deployments.get("MezoPortal")
+
+  const initialSupportedAssets = [
+    "0x7A56E1C57C7475CCf742a1832B028F0456652F97", // SolvBTC
+    "0xd9D920AA40f578ab794426F5C90F6C731D159DEf", // SolvBTC.BBN
+  ]
```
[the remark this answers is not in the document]
nkuba: We don't expect to add more tokens, and we don't need to handle testnet tokens or mock them, as we're testing on the main net fork and don't expect to plug this contract into the testnet dApp (at least for now).
Thank you for sharing your experience! I'll keep it in mind in case we want to do something more with this contract.

solidity/contracts/AcreMultiAssetVault.sol
jagodarybacka: > why Mezo Portal's withdraw function checks if the asset is supported

I think it was just for consistency.

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,279 @@
...
+            revert DepositNotFound();
+        }
+
+        // Withdraw the assets from the MezoPortal. Check if the withdrawn amount
+        // matches the deposited amount.
+        uint256 initialBalance = IERC20(asset).balanceOf(address(this));
+
+        // slither-disable-next-line reentrancy-no-eth
+        mezoPortal.withdraw(
+            address(asset),
```
pdyraga, line 237: This casting looks redundant.

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,279 @@
...
+
+    /// @notice Deposits an asset into the vault.
+    /// @param asset Address of the asset to be deposited.
+    /// @param amount Amount of the asset to be deposited (in the asset precision).
+    /// @return The ID of the deposit.
+    function deposit(address asset, uint256 amount) external returns (uint256) {
+        return depositFor(asset, amount, msg.sender);
+    }
+
+    /// @notice Deposits assets into the vault on behalf of a specified account.
```
pdyraga, line 158: It is worth adding to the docs only that specified account will be able to withdraw later.

pdyraga (commented):
Some minor comments to the contract code. Looking at the tests now.

solidity/deploy/41_deploy_multi_asset_vault.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const [_, deployment] = await helpers.upgrades.deployProxy(
```
pdyraga, line 16: Don't we want to check [if it was already deployed](https://github.com/thesis/mezo-portal/blob/24ebdee07b02e89de1db831d774bebc4319fd9bb/solidity/deploy/03_deploy_portal.ts#L12-L15)? Could save some pain with `func.skip` on various networks.

solidity/deploy/41_deploy_multi_asset_vault.ts
```diff
@@ -0,0 +1,46 @@
...
+    {
+      contractName: "AcreMultiAssetVault",
+      initializerArgs: [
+        await governance.getAddress(),
+        mezoPortal.address,
+        initialSupportedAssets,
+      ],
+      factoryOpts: { signer: deployer },
+      proxyOpts: {
+        kind: "transparent",
```
pdyraga, line 27: Assuming the new behavior of the deploy plugin is to create a separate `ProxyAdmin` for each contract (to be confirmed, I am never sure which behavior is old and which one is new), we should add an `initialOwner` parameter to `proxyOpts`.

>  The proxy admin contract also defines an owner address which has the rights to operate it. By default, the proxy admin’s owner is the initialOwner address used during deployment of the transparent proxy if provided, otherwise it is the externally owned account used during deployment. 

https://docs.openzeppelin.com/upgrades-plugins/1.x/

solidity/test/integration/Deployment.test.ts
```diff
@@ -236,4 +239,40 @@ describe("Deployment", () => {
...
+          `asset ${i}/${expectedSupportedAssets.length} is not supported: ${expectedSupportedAssets[i]}`,
+        ).to.be.true
+      }
+    })
+
+    it("should set owner", async () => {
+      expect(await multiAssetVault.owner()).to.be.equal(
+        expectedMainnetAddresses.governance,
+      )
+    })
```
pdyraga, line 276: Could be worth finding the ProxyAdmin and asserting its owner as well.

solidity/hardhat.config.ts
```diff
@@ -54,7 +54,7 @@ const config: HardhatUserConfig = {
       initialBaseFeePerGas: 1000000000,
       forking:
         process.env.FORKING === "true"
-          ? { url: MAINNET_RPC_URL, blockNumber: 20614139 }
+          ? { url: MAINNET_RPC_URL, blockNumber: 20937910 }
```
pdyraga, line 57: What about adding a code comment about what happened at that particular block?

solidity/test/integration/AcreMultiAssetVault.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+        let depositor: HardhatEthersSigner
+        let depositOwner: HardhatEthersSigner
+        let receiver: HardhatEthersSigner
+
+        before(async () => {
+          asset = await ethers.getContractAt("IERC20", assetAddress)
+
+          await impersonateAccount(assetHolderAddress)
+          await setBalance(assetHolderAddress, to1e18(1))
```
pdyraga, line 66: Not for this PR but I would consider renaming `setBalance` to `setEthBalance`. For a moment I was confused if this is Solv BTC balance and if so, why do we allow it to be lower than the deposit amount from the test data.

solidity/test/AcreMultiAssetVault.test.ts
```diff
@@ -0,0 +1,949 @@
...
+        .to.emit(multiAssetVault, "DepositCreated")
+        .withArgs(
+          await caller.getAddress(),
+          await asset1.getAddress(),
+          expectedDepositId,
+          depositAmount,
+        )
+    })
+
+    it("should transfer the asset to the vault", async () => {
```
pdyraga, line 257: Not really to the vault but to the Mezo Portal looking at `changeTokenBalances` assertion.

solidity/test/AcreMultiAssetVault.test.ts
```diff
@@ -0,0 +1,949 @@
...
+            .to.emit(multiAssetVault, "DepositCreated")
+            .withArgs(
+              await depositOwner.getAddress(),
+              await asset1.getAddress(),
+              expectedDepositId,
+              depositAmount,
+            )
+        })
+
+        it("should transfer the asset to the vault", async () => {
```
pdyraga, line 511: To the Mezo Portal.

solidity/test/AcreMultiAssetVault.test.ts
```diff
@@ -0,0 +1,949 @@
...
+      it("should withdraw correct deposit 3", async () => {
+        await testWithdrawDeposit(deposit3)
+      })
+
+      async function testWithdrawDeposit({
+        asset,
+        depositAmount: expectedWithdrawAmount,
+        depositOwner,
+        depositId,
+      }: Deposit & { depositId: bigint; receiver: HardhatEthersSigner }) {
```
pdyraga, line 854: I am confused, where do we pass the value for the `receiver`?

pdyraga: This entire construction is a bit twisted. I wonder if we can't just add `depositId` field to the `Deposit` struct.

pdyraga:
Added a bunch of comments but the only really semi-blocking item for me is [this one](https://github.com/thesis/acre/pull/762#discussion_r1796609135) about setting the `initialOwner` so that we do not have to worry about it post-deployment or even worse, forget.

solidity/test/integration/AcreMultiAssetVault.test.ts
nkuba: Agree, but this is a function imported from `@nomicfoundation/hardhat-toolbox/network-helpers`.
We can try to wrap it up with our hardhat helpers.

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,279 @@
...
+contract AcreMultiAssetVault is Ownable2StepUpgradeable {
+    using SafeERC20 for IERC20;
+
+    /// @notice Address of the MezoPortal contract.
+    IMezoPortal public mezoPortal;
+
+    /// @notice Reverts when a new asset is added to the list of supported assets.
+    error AssetAlreadySupported();
+
+    /// @notice Reverts when an asset is removed from the list of supported assets.
```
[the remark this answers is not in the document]
nkuba: 2d05a1746e27dc442bc2a614f87174dd96d8b8a9

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,279 @@
...
+    /// @param asset Address of the asset to be deposited.
+    /// @param amount Amount of the asset to be deposited (in the asset precision).
+    /// @return The ID of the deposit.
+    function deposit(address asset, uint256 amount) external returns (uint256) {
+        return depositFor(asset, amount, msg.sender);
+    }
+
+    /// @notice Deposits assets into the vault on behalf of a specified account.
+    /// @param asset Address of the asset to be deposited.
+    /// @param amount The amount of assets to deposit.
```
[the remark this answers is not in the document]
nkuba: 2d05a1746e27dc442bc2a614f87174dd96d8b8a9

solidity/contracts/AcreMultiAssetVault.sol
nkuba: 2d05a1746e27dc442bc2a614f87174dd96d8b8a9

solidity/contracts/AcreMultiAssetVault.sol
nkuba: 76218d756741202c62c33a649d4203fdcf8ecd14

solidity/deploy/41_deploy_multi_asset_vault.ts
nkuba: 4831b57462b9773fcdf8061b1d3617b754d33508

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,284 @@
...
+    mapping(address => mapping(address => mapping(uint256 => DepositInfo)))
+        public deposits;
+
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(
+        address _owner,
```
lukasz-zimnoch, line 100: Just flagging: Passing the `_owner` this way is a bit against the philosophy of the 2-step upgrade. We don't have any guarantee this `_owner` is correct and can manage the contract. The traditional approach with `__Ownable_init(msg.sender)` and explicit 2-step ownership transfer does not have this downside.

solidity/deploy/41_deploy_multi_asset_vault.ts
nkuba: Amazing catch! 🎉 
293ea548bd269a77448c7340d0a62149e3e9227e

solidity/test/integration/Deployment.test.ts
nkuba: f0f1a41b6ba8938f34c633bec323ffb703e9bf54

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,284 @@
...
+        uint256 amount,
+        address depositOwner
+    ) public returns (uint256) {
+        if (!supportedAssets[asset]) {
+            revert AssetNotSupported();
+        }
+        if (amount == 0) {
+            revert InvalidAmount(amount);
+        }
+        if (depositOwner == address(0) || depositOwner == asset) {
```
lukasz-zimnoch, line 180: It's still possible to use another supported asset as `depositOwner`. We may consider extending this check to:
```suggestion
        if (depositOwner == address(0) || supportedAssets[depositOwner]) {
```

solidity/hardhat.config.ts
nkuba: f069bf7902d615c531a53e22580d5324fdeb4861

solidity/test/AcreMultiAssetVault.test.ts
nkuba: 6bd6ff4df9709343674e7f6c96c2283b7fbcc80d

solidity/test/AcreMultiAssetVault.test.ts
nkuba: 6bd6ff4df9709343674e7f6c96c2283b7fbcc80d

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,284 @@
...
+
+        // Generate a new deposit ID.
+        uint256 depositId = ++depositCount;
+
+        emit DepositCreated(depositOwner, asset, depositId, amount);
+
+        // Transfer the asset from the deposit owner to the vault and approve
+        // Mezo Portal to pull them.
+        IERC20(asset).safeTransferFrom(msg.sender, address(this), amount);
+        IERC20(asset).forceApprove(address(mezoPortal), amount);
```
lukasz-zimnoch, line 192: There is a theoretical reentrancy risk here as we are interacting with a third-party address before making state changes. Sure, the asset must be supported here and in the Mezo Portal so we may consider the address as trusted but I would consider adding reentrancy protection just to reduce the potential attack surface if things go wrong in the future (e.g. a malicious token is added by mistake).

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,284 @@
...
+    /// @param asset Address of the asset to be withdrawn.
+    /// @param depositId ID of the deposit.
+    /// @param receiver Address of the account to which the assets will be sent.
+    /// @return The amount of assets withdrawn.
+    function withdraw(
+        address asset,
+        uint256 depositId,
+        address receiver
+    ) external returns (uint256) {
+        if (receiver == address(0) || receiver == asset) {
```
lukasz-zimnoch, line 223: It's still possible to use another supported asset as `receiver`. We may consider extending this check to:
```suggestion
        if (receiver == address(0) || supportedAssets[receiver]) {
```

solidity/contracts/AcreMultiAssetVault.sol
```diff
@@ -0,0 +1,284 @@
...
+
+        // Check if the deposit exists. If the deposit was not created or was
+        // already withdrawn, the depositedAmount will be 0.
+        if (depositedAmount == 0) {
+            revert DepositNotFound();
+        }
+
+        // Withdraw the assets from the MezoPortal. Check if the withdrawn amount
+        // matches the deposited amount.
+        uint256 initialBalance = IERC20(asset).balanceOf(address(this));
```
lukasz-zimnoch, line 238: Similar comment regarding the reentrancy risk as in `depositFor`.

solidity/contracts/AcreMultiAssetVault.sol
nkuba: Good point, I decided to introduce it after @pdyraga suggestion from https://github.com/thesis/acre/pull/762#discussion_r1795019308.

My reasoning is that we use the governance address that proved to work correctly.
What is more this is the deployment of the contract, so after we deploy the contract, we will check if the owner has been set to the correct address. If not, we will deploy another instance.

solidity/contracts/AcreMultiAssetVault.sol
nkuba: I initially thought about it, but then decided to use just the asset that is used in the call, as it's more likely user will confuse parameters of the function call and copy the same address twice, rather than they use another supported asset.
But since you raised that I included your suggestion in 690c2471a5d90efc3d50cda8ccccd43c07077ded.

solidity/contracts/AcreMultiAssetVault.sol
nkuba: 552e89fcf37874305ce0703a2effeccba57fcc2f

lukasz-zimnoch (dismissed):
I have reviewed only Solidity changes. LGTM!

Issue #650, opened 2024-07-22
no scroller in "connect your wallet" modal when necessary (season1 & /dashboard)


pnicjanik:
1. navigate to https://dapp.test.acre.fi/dashboard
2. change browser window resolution, reducing height modal starts shrinking >> options obfuscated after resize

takes a really (unrealistically) small screen/window resolution to do this now (sreenshot is 1280x505) ... but will be needed at larger resolutions after additions of XVerse & Ledger as connection options.

![acre_dash_connectWallet 2024-07-22 at 15 59 34](https://github.com/user-attachments/assets/dcc003d5-562a-4a9d-96ab-618a75f41862)

Issue #621, opened 2024-07-18
deposit received modal: items overlap on browser windows with smaller height resolutions


pnicjanik:
**prerequisite: browser height ~ 732px**

- https://dapp.test.acre.fi/dashboard
- connect wallet
- initiate & complete deposit
- modal content changes to "Deposit received" ...  "View on Mempool" link partially obfuscated by "Go to dashboard" button

![acre_deposit-received_20240718-173135](https://github.com/user-attachments/assets/d889faa6-9e0d-4a71-937b-84e5cd76a458)

Issue #757, opened 2024-10-03
Elements proportions & responsiveness


SorinQ:
Following our conversation on Discord, I’ve created this issue to address the problem with breakpoints and ensure responsiveness, even on the smallest screens.

https://discord.com/channels/1079490158639976609/1290310716620017736

I've put together a sample in Figma to help us with this. 
https://www.figma.com/design/lEtpjDlSQbAAyqwQdnDU0U/Ledger-dApp?node-id=13441-30475

@kkosiorowska Please let me know if anything else is required.

PS. Mobile version is also in progress and will be delivered separately.

nkuba:
@kkosiorowska has this issue been covered by https://github.com/thesis/acre/pull/766?

kkosiorowska:
> @kkosiorowska has this issue been covered by #766?

Yes, so I think we can close this task.

PR #766, opened 2024-10-15
Fix responsiveness on small window size for modal windows

kkosiorowska:
Closes https://github.com/thesis/acre/issues/650
Closes https://github.com/thesis/acre/issues/621
Ref https://github.com/thesis/acre/issues/757

This PR improves responsiveness on small window size for modal windows. What has been done:

- Set scrolling behavior for the modal outside
- Make the scrollbar visible when the sidebar is open.

  The sidebar has a higher z-index than the modal. When the scrollbar appears for the modal window it is covered by the sidebar. Therefore, we should check when the scrollbar is visible and then set the position of the Sidebar correctly.
  
  Note that we can not change the z-index for the Sidebar because the container for the modal window covers all window and will not allow a user to click on the element.

https://github.com/user-attachments/assets/7820e1ad-f0c2-426b-95d5-d113cbac36b5


r-czajkowski:
I noticed that I can't scroll when the sidebar element is focused. Also I can't use the scrollbar because it's only visible when I scroll.


https://github.com/user-attachments/assets/03c01657-adf6-49c3-97a2-af55eb914ed5

kkosiorowska:
> I noticed that I can't scroll when the sidebar element is focused. 

I honestly don't know how we can fix it and whether it can be done at all. 🤔 This is definitely a topic for more investigation.

>Also I can't use the scrollbar because it's only visible when I scroll.

I think it is caused by your system settings. But I noticed that when not displaying the scrollbar on the Mac the margin is not set correctly. But I think that we can't do anything about it.

Screenshot 2024-10-17 at 10 15 33

Anyway, I will create a task to look into it.

dapp/src/hooks/useScrollbarVisibility.ts
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useEffect, useState } from "react"
+
+const SCROLLBAR_WIDTH = `${window.innerWidth - document.body.offsetWidth}px`
+
+function scrollbarVisible(className: string) {
```
kpyszkowski, line 5: Non blocking:
It's a common pattern to select DOM elements with selectors, not only class names so I'd use `querySelector` instead of `getElementsByClassName`. Also the name made me confused. I had to double check what the function does. What do you think?

```suggestion
function isScrollbarVisible(selector: string) {
```

dapp/src/hooks/useScrollbarVisibility.ts
```diff
@@ -0,0 +1,29 @@
...
+      setIsVisible(scrollbarVisible(className))
+    }
+    window.addEventListener("resize", handleResize)
+
+    return () => {
+      window.removeEventListener("resize", handleResize)
+    }
+  }, [className])
+
+  const refetch = useCallback(() => {
```
kpyszkowski, line 24: Non blocking:
Fetching in computer science has numerous meanings. Generally speaking it describes the process of gathering data. The `refetch` name made me confused. The first what i look for when I see term `fetch` is the call of `axios` or `fetch` functions. I'd suggest to change it to something more descriptive like `handleScrollbarVisibilityCheck`. What do you think?

kpyszkowski (changes requested):
LGTM. Left 2 non blocking comments

> I noticed that I can't scroll when the sidebar element is focused.

It's about how CSS stacking contexts works. The Sidebar element is lower in DOM tree than scrollable area (modal container). By visually elevating the Sidebar over the modal  - with `z-index` property, interactions like scrolling are hijacked and are not propagated to the scrollable area.

I see two solutions: 

1. Render Sidebar component within modal portal so we can get rid of `z-index` property (and interactions hijacking).
2. Scroll document's `body` instead of modal container so it both modal and Sidebar will stay in the same stacking context. 

1. approach should be easier. Both solutions will make the scrollbar render over the content so it won't be covered by the Sidebar element (I believe so 😆)

dapp/src/hooks/useScrollbarVisibility.ts
kkosiorowska: https://github.com/thesis/acre/pull/766/commits/f75fe59747eac984f19eab26303094e32128b078 https://github.com/thesis/acre/pull/766/commits/038a7f2e27d883b8f2a2fdb28a8a4a0fc3764b4b

dapp/src/hooks/useScrollbarVisibility.ts
kkosiorowska: This function checks and sets the state again `isVisible`. So let's change its name simply to `refreshState`.

kpyszkowski (approved):
LGTM

PR #771, opened 2024-10-17
dApp: Fix ` $.getBitcoinProvider is not a function` error

kpyszkowski:
Sentry error [here](https://acre-e5.sentry.io/issues/5987899907/?project=4507545986793472&query=&referrer=issue-stream&sort=freq&statsPeriod=7d&stream_index=1)

My assumption is it occurs when user has deprecated version of the wallet extension. With explicit check if the function exists the issue is resolved.


kkosiorowska:
> My assumption is it occurs when user has deprecated version of the wallet extension. With explicit check if the function exists the issue is resolved.

I think that the version of the extension does not affect this error here.  Hook `useConnector` returns a connector from `OrangeKit` which should have an implemented method `getBitcoinProvider`. See [here](https://github.com/thesis/orangekit/blob/0519757bdd3f97c38cbdb07d2d3496c561a674bc/typescript/react/src/wallet/connector.ts#L185-L187
). Therefore, we do not need to check if the function exists in the connector object.

kkosiorowska (approved):
Let's not block this PR. But we should consider in the next steps why wagmi returns a connector without this method. More info [here](https://github.com/thesis/acre/pull/708#discussion_r1808895803).

Issue #789, opened 2024-10-28
Action duration estimation


kpyszkowski:
### Objective
To present the time estimation so user can be provided with the feedback of how long the action might take.

### Designs
[Figma](https://www.figma.com/design/lEtpjDlSQbAAyqwQdnDU0U/Ledger-dApp?node-id=14062-3762&t=TfbATDok9wfYNHdB-4)

<img width="993" alt="image" src="https://github.com/user-attachments/assets/ac738a8b-c107-40e7-bdf1-9adf87a19975">

### Dev notes 
The [`getEstimatedDuration`](https://github.com/thesis/acre/blob/main/dapp/src/utils/activities.ts#L23) function can be used.

```[tasklist]
### Tasks
- [x] Create component
- [x] Render the component within the action form accordingly
```

PR #790, opened 2024-10-28
dApp: Action duration estimation & validation

kpyszkowski:
Closes: #789 

### Changes:
- Implemented action duration estimation based on given amount
- Updated form validation strategy to update the state dynamically based on given value

https://github.com/user-attachments/assets/1893466c-910a-4933-8293-163da9c45f60


dapp/src/components/shared/TokenAmountForm/index.tsx
```diff
@@ -37,7 +37,7 @@ const TokenAmountForm = withFormik<TokenAmountFormProps, TokenAmountFormValues>(
       props.onSubmitForm(values)
     },
     validateOnBlur: false,
-    validateOnChange: false,
+    validateOnChange: true,
```
nkuba, line 40: We wanted to avoid running the validation on any input change for the amount field.
Currently, when you start entering the value starting with `0` you will immediately see the validation error. We want to run the validation when the user exits the input field.
So `validateOnBlur` seems like a better fit.

kkosiorowska: Yeah, I agree. Let's change it.

kpyszkowski: Ref commit: 878b1f43414632797ac00b8b3f7b6a405e6b1bb4

kkosiorowska (approved):

PR #797, opened 2024-10-29
Ledger Live fix connection errors

r-czajkowski:
Depends on: #780 

This PR fixes Ledger Live wallet connection bug. If the user rejects the login message signing and closes the modal, they should get back to the initial state of the not-connected account.

Steps to recreate the bug:
1. Open the account connection.
2. Select a supported account.
3. Cancel login message signing and clse the connection modal.
4. The dApp looks like it is connected to an account, but after switching the account, it shows a "Something went wrong" error.


nkuba (approved):
I tested the scenario I executed when the error occurred. It worked fine with this fix.

Issue #815, opened 2024-10-30
Update UI of "Withdrawal initiated!" screen


SorinQ:
As Is: The current UI of the Withdrawal Initiated screen is outdated and requires content updates and a style refresh.

To Be: Update the Withdrawal Initiated screen in two phases:

- Phase I: Update the content on the screen to ensure it is accurate and relevant. (mandatory until lunch)
- Phase II: Redesign the screen to match the latest UI style. (can be done after the launch)

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-3777&t=Y9P9w96lMLaQ5nbz-4

Issue #812, opened 2024-10-30
Update UI of "Deposit Received" Screen 


SorinQ:
As Is: The current UI of the Deposit Received screen is outdated and requires content updates and a style refresh.

To Be: Update the Deposit Received screen in two phases:

- Phase I: Update the content on the screen to ensure it is accurate and relevant. (mandatory until lunch)
- Phase II: Redesign the screen to match the latest UI style. (can be done after the launch)

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=10-28886&t=Y9P9w96lMLaQ5nbz-4

PR #820, opened 2024-10-30
Update UI of `SuccessModal`

kkosiorowska:
Closes https://github.com/thesis/acre/issues/812
Closes https://github.com/thesis/acre/issues/815


dapp/src/assets/icons/index.ts
```diff
@@ -2,7 +2,7 @@ export * from "./Info"
 export * from "./ArrowUpRight"
 export * from "./AcreLogo"
 export * from "./Pause"
-export * from "./LoadingSpinnerSuccessIcon"
```
kpyszkowski, line 5: Why did you change the type of export here? It brings inconsistency against other modules.

kpyszkowski (commented):
Left one comment:

dapp/src/assets/icons/index.ts
kkosiorowska: I changed the type because because we should use default exports. More info you can find [here](https://github.com/thesis/acre/commit/4b825290f930f3c934e879c8a08ebb693f0c7933#discussion_r1624438894).

I believe that soon we will have some time to update it everywhere.

r-czajkowski (approved):

Issue #822, opened 2024-10-30
dApp: Implement the TVL component


kpyszkowski:
### The objective:
To prepare component to display the total value locked.

### Designs: 
[Figma](https://www.figma.com/design/lEtpjDlSQbAAyqwQdnDU0U/Ledger-2023?node-id=13985-2)

<img width="850" alt="image" src="https://github.com/user-attachments/assets/6aa87e38-0384-43c6-9245-9551b11c63dd">

PR #823, opened 2024-10-30
dApp: `AcreTVLProgress` component

kpyszkowski:
Closes: #822

### Changes:
- Removed the use of term `season` - we don't have seasons any more
- Implemented `AcreTVLProgress` component
- Updated the TVL hook
- Updated styles of `Progress` component

### Preview:

image

### Note:
To start with latest layout changes I set the origin branch to [`ledger-live-updates`](https://github.com/thesis/acre/tree/ledger-live-updates)


dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+            amount={totalValueLocked.value}
+            currency="bitcoin"
+          />
+
+          <TextMd>Total value locked</TextMd>
+        </Grid>
+
+        <VStack sx={styles.progressWrapper}>
+          <HStack sx={styles.progressLabelsWrapper}>
+            {[50, 100, 150, 200, 250].map((value) => (
```
kkosiorowska, line 39: I think these values should depend on `TOTAL_VALUE_LOCKED_CAP`. A quick solution that comes to my mind:

```
export const useTVL = () => {

  ...
  const totalValue = bigIntToUserAmount(
    TOTAL_VALUE_LOCKED_CAP,
    BITCOIN.decimals,
  )

  return { progress, value, isCapExceeded, totalValue }
}

```

```
const UNIT = 50

const { totalValue } = totalValueLocked
const scale = Array.from({ length: totalValue / UNIT },(_, idx) => totalValue - idx * UNIT).reverse()
```

Let me know what you think and if you may have any other solution.

kkosiorowska (commented):
Looks awesome 😍 I left minor comments.

pnpm-lock.yaml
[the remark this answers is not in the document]
kpyszkowski: Those changes are only in file formatting. I use the latest version of pnpm - maybe they did some update to lockfile parser.

dapp/src/assets/icons/BoltFilled.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import { createIcon } from "@chakra-ui/react"
+
+export const BoltFilled = createIcon({
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 1953d7d0

dapp/src/assets/icons/index.ts
```diff
@@ -9,3 +9,4 @@ export * from "./MezoSignIcon"
 export * from "./AcreSignIcon"
 export * from "./BitcoinsStackErrorIcon"
 export { default as MatsIcon } from "./MatsIcon"
+export * from "./BoltFilled"
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 1953d7d0

dapp/src/hooks/index.ts
```diff
@@ -20,7 +20,7 @@ export * from "./usePagination"
 export * from "./useModal"
 export * from "./useTransactionModal"
 export * from "./useVerifyDepositAddress"
-export * from "./useSeasonProgress"
+export * from "./useTVL"
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: 1953d7d0

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
kpyszkowski: I discussed it with @nkuba.
The conclusion was to hardcode values.
Context [here](https://discord.com/channels/1079490158639976609/1297889075050582069/1300486504128446628)

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <Box sx={styles.container} {...props}>
+      <HStack sx={styles.wrapper}>
+        <Grid sx={styles.contentWrapper}>
+          <BitcoinIcon sx={styles.valueIcon} />
+
+          <CurrencyBalance
+            size="3xl"
+            amount={totalValueLocked.value}
```
nkuba, line 30: Currently, we show `244.6570`. Can we reduce the number of decimals? How about we show two decimals?

dapp/src/utils/acreApi.ts
```diff
@@ -106,6 +106,18 @@ async function getBtcUsdPrice(): Promise<number> {
   return response.data.btcUsdPrice
 }
 
+type GetTotalValueLockedResponse = {
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit:  a65743cb

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
kpyszkowski: Ref commit: 880130a9

kpyszkowski:
@nkuba 
As requested, updated the cap to 1250.
Ref commit: e586c639

dapp/src/constants/staking.ts
```diff
@@ -1,4 +1,4 @@
 // TODO: Read the value from the SDK, once we expose it
 export const MINIMUM_BALANCE = BigInt(String(5e6)) // 0.05 BTC
 
-export const SEASON_CAP = BigInt(String(5e10)) // 500 BTC
+export const TOTAL_VALUE_LOCKED_CAP = 1250 // 1250 BTC
```
r-czajkowski, line 4: Non-blocking comment: what do you think about passing this value as an env variable? We could change the cap w/o updating the code.

dapp/src/components/shared/ProgressBar.tsx
```diff
@@ -1,16 +1,28 @@
...
 
   return (
     <Progress value={value} hasStripe {...restProps}>
-      <ProgressLabel>
-        {children}
+      <ProgressLabel>{children}</ProgressLabel>
 
-        <Box>{value}%</Box>
-      </ProgressLabel>
+      {value > 2 && (
```
r-czajkowski, line 12: Why greater than `2`? This is a shared component so are we sure it will be ok for other use cases?

dapp/src/hooks/useTVL.ts
```diff
@@ -0,0 +1,33 @@
...
+    refetchInterval: REFETCH_INTERVAL_IN_MILLISECONDS,
+  })
+
+  const totalAssets = data?.btc ?? 0
+
+  const isCapExceeded = totalAssets > TOTAL_VALUE_LOCKED_CAP
+
+  const progress = Math.floor(
+    isCapExceeded ? 100 : (totalAssets / TOTAL_VALUE_LOCKED_CAP) * 100,
+  )
```
r-czajkowski, line 24:
```suggestion
  const progress =
    isCapExceeded ? 100 :  Math.floor((totalAssets / TOTAL_VALUE_LOCKED_CAP) * 100)
```

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { CurrencyBalance } from "#/components/shared/CurrencyBalance"
+import ProgressBar from "#/components/shared/ProgressBar"
+import { TextMd, TextXs } from "#/components/shared/Typography"
+import { TOTAL_VALUE_LOCKED_CAP } from "#/constants"
+
+type AcreTVLProgressProps = StackProps
+
+const STEP_COUNT = 5
+
+const getSteps = () =>
```
r-czajkowski, line 21: Can't it just be a const value?

```suggestion
const getSteps =
```

dapp/src/theme/Progress.ts
```diff
@@ -8,48 +8,54 @@ import { keyframes } from "@chakra-ui/react"
...
 const multiStyleConfig = createMultiStyleConfigHelpers(parts.keys)
 
-const filledStyle = defineStyle((props) => {
-  const { colorScheme: color, theme, isIndeterminate, hasStripe } = props
+const trackStyle = defineStyle((props) => {
+  const { isIndeterminate, hasStripe } = props
 
-  const bgColor = `${color}.400`
-  const stripeColor = "opacity.white.1"
+  const bgColor = "#3A3328"
```
r-czajkowski, line 14: Let's define this color. If it's only used in the `Progress` component we can declare a const value but outside of the function - for example at the beginning of this file.

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
nkuba: Let's assume we set the cap to 1250, so we have a step every 250. When we hit it and update the `TOTAL_VALUE_LOCKED_CAP` to 1600, the new progress bar will have a step every 320. So, the previous steps and the goal of 1250 are no longer displayed. This may be confusing as we don't maintain the historic goals.

I preferred hardcoding the steps, as depending on the change in the cap, we need to consider the steps to keep them consistent across updates.

We can keep the current automated calculation, but we need to be aware that we need to revisit the steps in case of an update in the future. It is worth adding a comment in the code.

nkuba: Let's assume we set the cap to 1250, so we have a step every 250. When we hit it and update the `TOTAL_VALUE_LOCKED_CAP` to 1600, the new progress bar will have a step every 320. So, the previous steps and the goal of 1250 are no longer displayed. This may be confusing as we don't maintain the historic goals.

I preferred hardcoding the steps, as depending on the change in the cap, we need to consider the steps to keep them consistent across updates.

We can keep the current automated calculation, but we need to be aware that we need to revisit the steps in case of an update in the future. It is worth adding a comment in the code.

dapp/src/constants/staking.ts
nkuba: We need to consider the steps when we update the cap. Please see https://github.com/thesis/acre/pull/823#discussion_r1827782305

If we change just the cap, without updating the steps, we won't maintain the historic goals.

dapp/src/hooks/useTVL.ts
```diff
@@ -0,0 +1,33 @@
...
+  TOTAL_VALUE_LOCKED_CAP,
+} from "#/constants"
+import { acreApi } from "#/utils"
+import { useQuery } from "@tanstack/react-query"
+
+const { acreKeys } = queryKeysFactory
+
+const useTVL = () => {
+  const { data } = useQuery({
+    queryKey: [...acreKeys.totalAssets()],
```
nkuba, line 13: Total assets is a name specific to the [stBTC vault](https://github.com/thesis/acre/blob/main/solidity/contracts/stBTC.sol). But now we include in the TVL not only the stBTC vault total assets, but also deposits made to the [Multi-Asset Vault](https://github.com/thesis/acre/blob/main/solidity/contracts/AcreMultiAssetVault.sol).

Let's replace `totalAssets` with `tvlBtc` across the file. Let's also update the query keys factory.


```suggestion
    queryKey: [...acreKeys.tvlBtc()],
```

dapp/src/constants/staking.ts
kpyszkowski: Discussed with Kuba. Decided to hardcode it as a constant. It won't be changed that often.

dapp/src/hooks/useTVL.ts
kpyszkowski: Refactored for consistency: f4eb4bd1
You called the endpoint "statistics" so I followed the convention.

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
kpyszkowski: Ref commit: e985c20c

dapp/src/hooks/useTVL.ts
kpyszkowski: Ref commit: 6d9e39ba

dapp/src/components/shared/ProgressBar.tsx
kpyszkowski: Ref commit: c51c695c

r-czajkowski (approved):
LGTM! I think we can hold off on merging until we merge the https://github.com/thesis/acre/pull/708 into `main`.

dapp/src/pages/DashboardPage/AcreTVLProgress.tsx
kpyszkowski: The number of steps on the progress bar is currently parametrized with a variable.

```ts
const STEP_COUNT = 5
```
- 4 steps: image

- 5 steps: image

- 6 steps: image

In my opinion it's enough control for now.  We can adjust it later in case more control is needed (like irregular steps you mentioned).

nkuba (approved):

Issue #806, opened 2024-10-29
Update Info Tooltip for Beehive Component


SorinQ:
As Is: Currently, there is no additional information for the Beehive component, leading to confusion among users who may not realize that the displayed amount represents the total points, and they only hold a share of it.

To Be: Update info displayed inside toolip.

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-52678&t=jJPb7IQTImT8xBpU-4

Issue #799, opened 2024-10-29
Add Info Tooltip on Hover for Points


SorinQ:
As Is: Currently, there is no additional information provided for "Points". We need to explain to users that unclaimed points are not lost.

To Be: Add an info icon next to "Points" that, when hovered over, displays a tooltip with additional information about Points.

**Important: We have differente versions for Guest and User.**

Acceptance Criteria:

-     Info icon appears next to the Points label.
-     On hover, a tooltip displays, providing a brief explanation of Points.
-     Tooltip should be positioned clearly and be easily readable.

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-52003&t=jJPb7IQTImT8xBpU-4

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-52673&t=jJPb7IQTImT8xBpU-4

Issue #795, opened 2024-10-29
Display Tooltip with Full Amount on Hover


SorinQ:
As Is: When an amount is too long, it’s cut off with an ellipsis ("...") at the end.

To Be: Show a tooltip with the exact, full amount when hovering over any amount that displays an ellipsis.

Acceptance Criteria:

-     Tooltip appears on hover, showing the full amount without any truncation.
-     Tooltip should be clear and positioned close to the truncated amount for easy reading.

Sample:
https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-52683&t=jJPb7IQTImT8xBpU-4

PR #825, opened 2024-11-04
dApp: Tooltip updates

kpyszkowski:
Closes: #795
Closes: #799 
Closes: #806

### Changes
<details>
  <summary>Added `AcrePointsCard` component's conditional tooltip (depends if wallet is connected) </summary>
  image
  image
</details>
<details>
  <summary>Added `BeehiveCard` component's tooltip</summary>
  image
</details>
<details>
  <summary>Added `withToolip` prop to `CurrencyBalance` component which renders full (not truncated) amount of the activity</summary>
  image
</details>

### Note
UI changes (colors, sizes) are not included. This PR includes content/functionality changes only.


dapp/src/pages/DashboardPage/AcrePointsCard.tsx
```diff
@@ -47,10 +51,19 @@ export default function AcrePointsCard(props: CardProps) {
...
+      <CardHeader p={0} mb={2} as={HStack} justify="space-between">
         <TextMd fontWeight="bold" color="grey.700">
           Your Acre points balance
         </TextMd>
+
+        <InfoTooltip
+          label={
+            isWalletConnected
+              ? "Your current balance of Acre points collected so far. New points drop daily and are ready to be claimed. Unclaimed points roll over to the next day."
+              : "Total points distributed to Acre users so far. New points drop daily and can be claimed in each user's dashboard."
```
nkuba, line 63: The component does not show the total points when the user is not connected. We cannot add this tooltip variant unless we update the way we display points for non-connected wallets.

kpyszkowski: Good catch. I removed the variant for disconnected wallet.
The tooltip with appropriate message appear as the wallet is connected.
Left TODO comment.
Ref commit: 6e293bdd 8d531a65

dapp/src/components/shared/CurrencyBalance/index.tsx
```diff
@@ -76,11 +80,23 @@ export function CurrencyBalance({
...
+        {withDots && "..."}
       </Box>
       <Box as="span" __css={styles.symbol} {...textProps} {...symbolTextProps}>
         {symbol}
       </Box>
     </Box>
   )
+
+  if (withTooltip) {
+    const tooltipLabel = `${bigIntToUserAmount(BigInt(amount ?? 0), decimals, decimals)} ${symbol}`
```
kkosiorowska, line 92: We do not display the correct amount in the tooltip, see below: 

Screenshot 2024-11-05 at 14 49 46

Note that we actually want to calculate the value the same way as for the `balance` variable but in this case `desiredDecimals` will be equal to `decimals` from the currency.

kpyszkowski: > `desiredDecimals` will be equal to `decimals`

It's not necessarily the case. For Bitcoin the `desiredDecimals` is 8. When amount "length" is lower than that it causes the offset you noticed. It's impossible to define it with constant value since it's specific to the amount. Sometimes it's 6, 7, rarely 8.

```ts
const fullPrecisionDecimals = (amount ?? 0).toString().length
console.log(amount, fullPrecisionDecimals)
```

image

I came up with idea to dynamically read the precision per amount. Looks like it's accurate.

image

Also I transformed the `balance` memoized value to `getBalance` memoized function. It allowed me to get exact the same balance in multiple places without repetition of the code. Performance-wise it's also covered by memoization.

Ref commit: 0fe9988a

dapp/src/pages/DashboardPage/AcrePointsCard.tsx
```diff
@@ -29,6 +30,9 @@ export default function AcrePointsCard(props: CardProps) {
     updatePointsData,
     isCalculationInProgress,
   } = useAcrePoints()
+  const { address } = useWallet()
+
+  const isWalletConnected = !!address
```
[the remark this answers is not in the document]
kpyszkowski: Ref commit: ccf0521a

dapp/src/components/shared/CurrencyBalance/index.tsx
kpyszkowski: If the precision is critical and must be as accurate as possible then we use the method - `fixedPointNumberToString`. Inside the `bigIntToUserAmount` the `fixedPointDecimals` is subtracted by `desiredDecimals` which results as **0** (8 - 8). It leads to unexpected value offsets like **0.12200000** instead of expected **1.22000000**.

Ref commit: 9d13616c

kkosiorowska (approved):

Issue #805, opened 2024-10-29
[TVL]  Display "Remaining Until Deposit Cap" Next to Deposit Button


SorinQ:
As Is: Currently, there is no fast indication of the remaining amount until reaching the deposit cap near the Deposit button.

To Be: Add a label next to the Deposit button displaying the text "Remaining Until Deposit Cap" to show users how much deposit capacity remains.

Acceptance Criteria:

-"<value> BTC remaining until deposit cap" text appears next to the Deposit button.
-Displayed amount should dynamically update based on remaining deposit capacity.
- Ensure the label is styled consistently with the rest of the UI and is easily readable.

Additional Notes:
-Confirm that the remaining amount calculation is accurate and updates in real time, if applicable.

https://www.figma.com/design/Ov4M1oVlSrO4Pl7NmGOHC1/%22Zero%22?node-id=58-51511&t=jJPb7IQTImT8xBpU-4

PR #826, opened 2024-11-04
dApp: Deposit limit when TVL cap is exceeded

kpyszkowski:
Closes: #822

### Changes:
- Add information about the TVL cap excess
- Disable the deposit button when the TVL cap is exceeded

### Preview:

image

### Note:
To start with latest TVL changes I set the origin branch to [`tvl-component`](https://github.com/thesis/acre/tree/tvl-component)


dapp/src/pages/DashboardPage/PositionDetails.tsx
```diff
@@ -81,24 +89,33 @@ export default function PositionDetails() {
...
+                {...buttonStyles}
+                onClick={openWithdrawModal}
+                isDisabled={!isWithdrawalFlowEnabled}
+              >
+                Withdraw
+              </Button>
+            </ArrivingSoonTooltip>
+          </UserDataSkeleton>
+        )}
+        <AcreTVLMessage />
```
kkosiorowska, line 118: Shouldn't we hide it behind the `TVL_ENABLED` feature flag? 

cc @nkuba

dapp/src/pages/DashboardPage/PositionDetails.tsx
```diff
@@ -39,6 +42,11 @@ export default function PositionDetails() {
   const openWithdrawModal = useTransactionModal(ACTION_FLOW_TYPES.UNSTAKE)
   const activitiesCount = useAllActivitiesCount()
 
+  const { tvl } = useStatistics()
+
+  const { address } = useWallet()
+  const isWalletConnected = !!address
```
kkosiorowska, line 48: We can use here `isConnected` from `useWallet`.

dapp/src/hooks/useStatistics.ts
```diff
@@ -24,12 +24,15 @@ const useStatistics = () => {
     ? 100
     : Math.floor((bitcoinTvl / TOTAL_VALUE_LOCKED_CAP) * 100)
 
+  const remaining = TOTAL_VALUE_LOCKED_CAP - bitcoinTvl
```
kkosiorowska, line 27: I think we should not allow `remaining` to have a negative value here.

```suggestion
  const remaining = isCapExceeded ? 0 : TOTAL_VALUE_LOCKED_CAP - bitcoinTvl
```

dapp/src/pages/DashboardPage/PositionDetails.tsx
nkuba: That's a good catch. If we have the TVL top component configurable with a feature flag, we should also include the message and button locking.

dapp/src/hooks/useStatistics.ts
kpyszkowski: Ref commit: cfae577e

dapp/src/pages/DashboardPage/PositionDetails.tsx
kpyszkowski: Ref commit: cfae577e

kkosiorowska (commented):
LGTM, I left only one question and non-blocking comment.

- When the flag is disabled and the user is not connected,  we do not display the withdraw button. Currently we always display two buttons. Are we sure we want to change it?

Screenshot 2024-11-07 at 14 29 34

- I think we could improve a bit the loading of elements in the following component. What I mean is that when the user is not connected can see a button and text then later the text disappears and two buttons appear. I think we could add maybe a skeleton for all this row?

https://github.com/user-attachments/assets/9288aeb1-6932-4607-ac27-78ba7f4de230

nkuba:
@kpyszkowski does this PR address https://github.com/thesis/acre/issues/805?

kkosiorowska (approved):

PR #830, opened 2024-11-05
Fix wallet connection bug

r-czajkowski:
This PR fixes Xverse and Unisat connection bugs. We noticed problems with connecting Xverse and Unisat wallets in the dApp. Users are redirected to the Chrome store even though they have the extension installed. It started happening after I initially tried to connect with an invalid network selected in the wallet. The bug is that the `window.unisat` and `window.XverseProviders.BitcoinProvider` have `undefined` values. The orangekit package refers to these fields under the hood and the `isInstalled` function always returns `false`.

To fix this bug here we use the dynamic import to get functions that create orangekit connectors.

NOTE: 
I noticed different behavior in development and production builds. In the production build `window.unisat` and `window.XverseProviders.BitcoinProvider` are `undefined` even though they are installed. I was able to reproduce the bug in production on `stake.acre.fi`, but locally I was only able to do it after building the dapp (`pnpm run build`) and running it using `pnpm run preview`.


dapp/src/DApp.tsx
```diff
@@ -32,24 +33,45 @@ function DApp() {
 }
 
 function DAppProviders() {
+  const [config, setConfig] = useState<Config | undefined>()
+
+  useEffect(() => {
+    const setWagmiConfig = async () => {
+      await delay(500)
```
kpyszkowski, line 40: Non blocking:
Such time based approach seems prone to cause unexpected issues. 
What if 500ms is too little? Have you tried [asynchronous dynamic imports](https://vite.dev/guide/features#dynamic-import) for this issue? 

Looks like it might work here better.

kpyszkowski (commented):
Looks good to me. 
However in my opinion it looks more like workaround rather than solution.
I think it requires more investigation like dynamic imports I proposed.

dapp/src/DApp.tsx
r-czajkowski: The delay is only to display the "loading" screen for a bit longer than a quick flash so that it is noticeable for a moment. See the `getWagmiConfig` function. It uses dynamic import.

dapp/src/wagmiConfig.ts
```diff
@@ -21,21 +16,29 @@ const transports = chains.reduce(
...
 
-const orangeKitUnisatConnector = getOrangeKitUnisatConnector(connectorConfig)
-const orangeKitOKXConnector = getOrangeKitOKXConnector(connectorConfig)
-const orangeKitXverseConnector = getOrangeKitXverseConnector(connectorConfig)
+async function getWagmiConfig() {
+  const {
+    getOrangeKitUnisatConnector,
+    getOrangeKitOKXConnector,
+    getOrangeKitXverseConnector,
+  } = await import("@orangekit/react")
```
nkuba, line 24: Wouldn't it work if we keep the import static as we used to? Maybe wrapping the loading with `getWagmiConfig` function is enough to solve the problem?

nkuba (approved):
I left two comments. Please address them in a follow-up PR. I don't want to block this fix as it improves the production.

Issue #741, opened 2024-09-26
withdrawing less than allowed limit throws error that references deposit limit


pnicjanik:
reproduced in testnet (https://deploy-preview-704--acre-dapp-testnet.netlify.app/) and prod/mainnet (stake.acre.fi)

- logged in as a user with staked deposits
- click to `Withdraw`
- enter amount= `0.00001` & click to `Withdraw` 
- error thrown: "The amount is below the minimum required **deposit** of 0.01 BTC."

<img src="https://github.com/user-attachments/assets/1e70ff3f-591f-4b00-a68d-2e8b55174d91" width=404>

pnicjanik:
this issue's fix introduced: https://github.com/thesis/acre/issues/855

PR #833, opened 2024-11-07
Fix error message for withdraw form

kkosiorowska:
Closes https://github.com/thesis/acre/issues/741

Previously, a withdrawal less than the allowed limit caused an error that referred to the deposit limit. This PR in the error message sets the correct form type.

### UI

**Before**
Screenshot 2024-11-07 at 08 45 06

**After**
Screenshot 2024-11-07 at 08 37 34

Screenshot 2024-11-07 at 08 38 03


r-czajkowski (approved):

pnicjanik:
created https://github.com/thesis/acre/issues/855 - it should read "withdrawal" not "withdraw"

PR #838, opened 2024-11-08
Refactor for the card theme

kkosiorowska:
This PR standardizes our styles for the `Card` component. What has been done:

- In many places we have set `padding={0}` for the `CardBody`. Let's set a single common padding for the Card component.
- Updated header styles for the cards.
- Removed the border for the `Card` component and the unneeded `elevated` variant.
- Removed unneeded props that are already defined in the card theme.

⚠️ Important notice ⚠️
Many things probably still need to be improved, but I believe this PR cleans our code a bit. Let's make the next fixes and integrate with styleguide in the next steps.


dapp/src/theme/Card.ts
```diff
@@ -1,40 +1,29 @@
...
-  },
-  defaultProps: {
-    size: "lg",
-  },
-}
+const baseStyleContainer = defineStyle({
+  boxShadow: "none",
+  bg: "gold.200",
+  borderRadius: "xl",
+  padding: 5,
```
kpyszkowski, line 10: Please use shorthand instead
```suggestion
  p: 5,
```

dapp/src/theme/Card.ts
```diff
@@ -1,40 +1,29 @@
...
+})
+
+const baseStyleHeader = defineStyle({
+  padding: 0,
```
kpyszkowski, line 14: Same as above

dapp/src/theme/Card.ts
```diff
@@ -1,40 +1,29 @@
...
+})
+
+const baseStyleBody = defineStyle({
+  padding: 0,
```
kpyszkowski, line 18: Same as above

kpyszkowski (commented):
Left non blocking comments

dapp/src/theme/Card.ts
kkosiorowska: https://github.com/thesis/acre/pull/838/commits/dc231b83adad68ad3c5c2182854f0598fc36c11d

dapp/src/theme/Card.ts
kkosiorowska: https://github.com/thesis/acre/pull/838/commits/dc231b83adad68ad3c5c2182854f0598fc36c11d

kpyszkowski (approved):

PR #845, opened 2024-11-12
Update `signWithdrawMessage` in `AcreLedgerLiveBitcoinProvider`

r-czajkowski:
Update `signWithdrawMessage` method - based on the Acre module code we should pass the nonce as a hex string instead of the decimal value as string.

Ref: https://github.com/blooo-io/hw-app-acre/blob/main/src/AcreBtcNew.ts#L357
Ref: https://github.com/blooo-io/hw-app-acre/blob/main/src/AcreBtcNew.ts#L318


dapp/src/utils/orangekit/ledger-live/bitcoin-provider.ts
```diff
@@ -274,7 +274,7 @@ export default class AcreLedgerLiveBitcoinProvider
       message: {
         ...data,
         operation: data.operation.toString(),
-        nonce: data.nonce.toString(),
+        nonce: data.nonce.toString(16),
```
nkuba, line 277: Does it add `0x` prefix?

nkuba: Regardless of what the underlying library does, we should output the value in an unambiguous format. This is especially true here since one of the numbers (`operation`) is handled as a decimal string and another (`nonce`) as a hexadecimal.

Maybe we could use `ethers.utils.hexlify`?

r-czajkowski: Should we pass just converted number to hex with `0x` prefix, or add leading zeros? For example `12`:
1. `0xc`
2. `0x0c`
3.  `0x000000000000000000000000000000000000000000000000000000000000000c`

r-czajkowski: 925fc87

nkuba (approved):

PR #852, opened 2024-11-13
Make sure the connector is updated after every address change 

kkosiorowska:
Previously, the connector depended on the status from  `useAccount`.  Let's make sure the connector is updated after every address change.

**Steps to reproduce the issue:**

1. Log in to dApp with Unisat.
2. Confirm the connection in the Unisat window.
3. Close the modal window in the dApp immediately with the esc key


kkosiorowska:
PR is out of date.

Issue #867, opened 2024-11-15
We should estimate points on the deposit form based on the "You will deposit" amount


michalinacienciala:
Currently we are estimating points on the deposit form based on the deposit amount entered by the user. As [discussed on Discord](https://discord.com/channels/1079490158639976609/1306530286607273995/1306924401098227855), we think the points should be calculated on the value we estimate the deposit will have after fees subtraction (i.e. "You will deposit" amount).

PR #868, opened 2024-11-15
 Calculate points based on value after subtracting fees

kkosiorowska:
Closes https://github.com/thesis/acre/issues/867

We are estimating points on the deposit form based on the deposit amount entered by the user. we think the points should be calculated on the value we estimate the deposit will have after fees subtraction (i.e. "You will deposit" amount).

**Before**
Screenshot 2024-11-15 at 12 28 53

**After**

Screenshot 2024-11-15 at 12 37 13


r-czajkowski (dismissed):

kkosiorowska:
> You will deposit shows 0.139 BTC, points per day show +140. Shouldn't it be +139? Or is it justified by different handling of rounding in the points calculation script and in the You will deposit calculation?

Yeah, the difference is rounding. @nkuba how should we round the number? Should the user see the number 140 or 139? Should we round the number of points up or down?

nkuba:
@kkosiorowska Round down, please.

kkosiorowska:
> @kkosiorowska Round down, please.

Done - https://github.com/thesis/acre/pull/868/commits/fdd748ff49dde1c81e2dd3ab5a384fcb30f3e65b

Before
Screenshot 2024-11-18 at 16 25 08


After

Screenshot 2024-11-18 at 16 31 31

michalinacienciala (approved):

r-czajkowski (approved):

Issue #747, opened 2024-10-01
[Analytics] PostHog integration


kpyszkowski:
### Objective

The objective is to track usage, learn from actual user behavior to improve the product.

Things we care about: 
* Conversion funnel: Discovery > Deposit > Return
* Drop-off rate / exit points
* Traffic origin / attribution
* User data
  * Wallet type
  * BTC balance
  * Geolocation
* Sessions: overall, count per day, per week etc.
* Session Length: The duration of time users spend on the interface in one session.
* Task Success Rate
* Error Rate
* Time on Task: The time it takes for users to complete a task. (Deposit, Withdraw)

### Solution 

The solution is to integrate PostHog:
* Purpose: Product analytics.
* Importance: Provides detailed insights into user behavior and product usage, helping to optimize the user experience and guide data-driven product development.\
* How: [PostHog JS SDK](https://posthog.com/docs/libraries/node) and [Threshold Dashboard](https://github.com/threshold-network/token-dashboard) as a reference.

_\* Based on [product pitch document](https://coda.io/d/Acre_d3Vs1FswTQA/Analytics-setup_sucOf0DQ#_lujwPTEG) and [open discussion](https://discord.com/channels/1079490158639976609/1257846488248356904)._

```[tasklist]
### Tasks
- [ ] Install and configure `posthog-js` library: configuration, feature flag
- [ ] Implement basic hooks: identify, capture, capture page view
- [ ] Attach PostHog's captures to specific user behaviors: TBA
```

PR #871, opened 2024-11-16
dApp: PostHog integration

kpyszkowski:
Closes: #747
Closes: AENG-30

### Overview
This PR integrates dApp with PostHog service to track users activities and analyze it. The PostHog project was created. All dApp activity is now accessible through [the dashboard](https://us.posthog.com/project/104474).
The API key and endpoint are available in the project settings.

### Features
- dApp integration using [PostHog's official React SDK](https://www.npmjs.com/package/posthog-js)
- Autocapture: all user interactions are tracked, any button click, any form fill
- Manually triggered `$pageview` capture event: autocapture for page view event was disabled to prevent false positive captures. It is triggered manually as user visits the dApp.
- User identification via wallet address
- Added custom events with additional payload information: 
  - `deposit_success` - payload: `{ transactionHash }`
  - `deposit_failure` - payload: `{ cause }`
  - `withdrawal_success` - wasn't able to easily get any reasonable payload here, can be done later
  - `withdrawal_failure` - payload: `{ cause }`
  - `points_claim_success` - payload: `{ claimedAmount, totalAmount }`
  - `points_claim_failure` - payload: `{ cause }`
- Defined feature flag to toggle PostHog integration

### Changes
- Added new env variables to define API key and endpoint and a feature flag
- Defined preconfigured `PostHogProvider` to inject the service into application
- Defined hooks: `usePostHogCapture`, `usePostHogIdentity`, `usePostHogPageViewCapture`
- Integrated key actions to capture PostHog custom events (see above)

### Caveats
To ensure the `$pageview` event is captured properly it's needed to call it as a side effect as the `location` value changes. To access this value (via `useLocation` hook) the router's context needs to be in scope of PostHog's context.
The only way to achieve it was to do it in the `Layout` component where both contexts are available.

### Note
It's impossible to exclude captures for development environment. To keep data logs clean and ensure it contains only user's data I suggest to **keep the feature flag disabled** while using development environment.


dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx
```diff
@@ -63,8 +69,16 @@ export default function DepositBTCModal() {
...
         onError(error)
       }
+
+      const captureParameters =
+        error instanceof Error
+          ? {
+              cause: error.message,
+            }
+          : undefined
+      handleCapture(PostHogEvent.DepositFailure, captureParameters)
```
r-czajkowski, line 79: Maybe we should extract this code to a separate hook or util function? We do the same in `SignMessageModal.tsx` file.

kpyszkowski: Nice one
Ref commit: fa12d8a

kpyszkowski:
@r-czajkowski 

> `withdrawal_success` - wasn't able to easily get any reasonable payload here, can be done later

Do you have any ideas about payload we could attach to the capture there? I mean something useful and available in the scope so can be used.

r-czajkowski:
> @r-czajkowski
> 
> > `withdrawal_success` - wasn't able to easily get any reasonable payload here, can be done later
> 
> Do you have any ideas about payload we could attach to the capture there? I mean something useful and available in the scope so can be used.

Maybe we can add `transactionHash` and `redemptionKey` see https://github.com/thesis/acre/blob/main/sdk/src/modules/account.ts#L200.

kpyszkowski:
@r-czajkowski 
> Maybe we can add `transactionHash` and `redemptionKey` 

It's not that straight forward. I need to play around with mutation function to pass these parameters to callbacks.
Such change might affect whole withdrawal flow. I don't want to risk it. It's doable but let's not block the PR.

I prepared POC which looks fine for further investigation:
[withdrawal-success-payload-poc.patch](https://github.com/user-attachments/files/17950424/withdrawal-success-payload-poc.patch)

r-czajkowski (approved):
Let's remember to set up correct env variables on testnet and production. I belive we should disable the posthog support on testnet.

kpyszkowski:
Updated environmental variables:

<details>
  <summary>Mainnet (bitcoin.acre.fi)</summary>
  image
</details>

<details>
  <summary>Testnet (bitcoin.test.acre.fi)</summary>
  image
</details>

r-czajkowski (approved):

Issue #863, opened 2024-11-14
The transaction sign window appears despite the rejection of the deposit


kkosiorowska:
Steps to reproduce:
1. Go to deposit form and request deposit
2. Close `Opening your wallet for signature` screen
3. Wait

https://github.com/user-attachments/assets/8b50e3d0-0d1a-4b97-8f88-2a2e2c28bcb1

PR #877, opened 2024-11-19
Don't open the wallet window after closing deposit flow

kkosiorowska:
Closes https://github.com/thesis/acre/issues/863

The transaction sign window appears despite the rejection of the deposit.  Let's already use the solution for the [withdrawal](https://github.com/thesis/acre/pull/670) flow. The cancel logic was moved to `useCancelPromise`.

https://github.com/user-attachments/assets/76219622-850c-4aeb-b5cb-f2b44aa884e4


dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx
```diff
@@ -28,6 +29,12 @@ export default function DepositBTCModal() {
     queryKey: userKeys.balance(),
   })
 
+  const sessionId = useRef(Math.random())
+  const { cancel, resolve, sessionIdToPromise } = useCancelPromise(
+    sessionId.current,
+    "Deposit cancelled",
+  )
```
r-czajkowski, line 36: Maybe we can generate sessionId inside `useCancelPromise` hook?

dapp/src/hooks/useCancelPromise.ts
r-czajkowski, line 1: Looks like we can hide the `sessionIdToPromise` in `useCancelPromise` hook and generate the session id here. For example:
`useCancelPromise.ts`
```ts
const sessionId = ... //generate session here

... // Implementation here
 
return {
 cancel,
 resolve,
 options: { shouldOpenErrorModal: false  } 
} 
```

dapp/src/components/TransactionModal/ActiveStakingStep/DepositBTCModal.tsx
kkosiorowska: Previously I used this solution but had the problem that `shouldOpenErrorModal` was not returning the correct value for the current `sessionId`. 

Steps to reproduce:
1. Reject the withdrawal request.
2. Try again to do the withdrawal.

The problem was that the wallet window appeared and the dApp  displayed a modal error. This happened because `shouldOpenErrorModal` was true. Hook didn't return the right value for the session.

Screenshot 2024-11-26 at 15 13 29

Screenshot 2024-11-26 at 15 24 42

I decided to pass `sessionId` to the hook to make sure we get the right value for `shouldOpenErrorModal`.  This solution solved the problem.

r-czajkowski (approved):

r-czajkowski (approved):

Issue #876, opened 2024-11-19
`acre.referral` set to `0` for Ledger Live after clearing local storage


michalinacienciala:
Steps to reproduce:
1. While logged in or logged out in LL Acre clear local storage
2. Click Refresh in LL

![image](https://github.com/user-attachments/assets/bf92bd0a-6e87-4254-81b1-6e16efaac300)

The `acre.referral` key should always be set to `2083` value for users using Acre in Ledger Live.

PR #886, opened 2024-11-20
Update the referral mechanism

r-czajkowski:
Closes: #876 

Detect the referral code in the main router loader instead of the hooks. Using a router loader is simpler and we avoid race between two hooks `useDetectEmbed` and `useDetectReferral`.

We noticed a bug with hooks implementation - `acre.referral` set to `0` for Ledger Live after clearing local storage. The `acre.referral` key should always be set to `2083` value for users using Acre in Ledger Live.


kkosiorowska (commented):
Overall it looks good. I have only one comment.

I tested the referral in the standalone dApp, then removed it from URL. The value has not been reset.  I think that in such a case we should back to the default value.
Screenshot 2024-11-21 at 14 29 01

Screenshot 2024-11-21 at 14 29 21

r-czajkowski:
> Overall it looks good. I have only one comment.
> 
> I tested the referral in the standalone dApp, then removed it from URL. The value has not been reset. I think that in such a case we should back to the default value. Screenshot 2024-11-21 at 14 29 01
> Screenshot 2024-11-21 at 14 29 21

[ac31173](https://github.com/thesis/acre/pull/886/commits/ac311737feabe9bdf4ef4b347b8bd40a28d9e8e2)

dapp/src/router/index.tsx
```diff
@@ -32,6 +34,24 @@ export const router = createBrowserRouter([
   {
     path: "/",
     element: <Layout />,
+    loader: ({ request }) => {
+      // TODO: display the error page/modal when the referral is invalid.
```
kkosiorowska, line 38: There is a case where we have a correct `ref` value in the URL, but then change it to an incorrect one. We still store the previous value in `localStorge`.  Shouldn't we also reset this value? 

Screenshot 2024-11-22 at 09 01 19

r-czajkowski: aafe02f

kkosiorowska (approved):

PR #903, opened 2024-11-27
Log connection error

nkuba:
We are now logging connection errors to the console. This will help us debug connection issues, as the error will be reported to Sentry.

We're handling only the `DEFAULT` error type, as the other types are reported to the user in a modal.


kkosiorowska (approved):
LGTM

PR #904, opened 2024-11-27
Set Sentry user id to hashed identifier

nkuba:
We calculate the ID of the user by hashing the address. This way we won't store any sensitive information in Sentry.

### DIFF
--- a/dapp/package.json
+++ b/dapp/package.json
@@ -35,7 +35,6 @@
     "@tanstack/react-query-devtools": "^5.49.2",
     "axios": "^1.6.7",
     "bignumber.js": "^9.1.2",
-    "crypto-js": "^4.2.0",
     "ethers": "^6.10.0",
     "formik": "^2.4.5",
     "framer-motion": "^10.16.5",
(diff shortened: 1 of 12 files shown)


dapp/src/sentry.test.ts
kkosiorowska, line 1: We keep the tests in separate directories. So I'm wondering whether to move them to the [tests](https://github.com/thesis/acre/tree/main/dapp/src/tests) folder.

nkuba: I updated it in https://github.com/thesis/acre/pull/906 to consistently keep the test in the root `test` folder with other packages.
I propose we merge this PR and update it in #906.

dapp/package.json
```diff
@@ -35,6 +35,7 @@
     "@tanstack/react-query-devtools": "^5.49.2",
     "axios": "^1.6.7",
     "bignumber.js": "^9.1.2",
+    "crypto-js": "^4.2.0",
```
r-czajkowski, line 38: Do we need the new lib? I think we can use [ethers](https://docs.ethers.org/v6/api/crypto/#sha256) or [viem](https://viem.sh/docs/utilities/sha256.html).

nkuba: [679db3f](https://github.com/thesis/acre/pull/904/commits/679db3f05a26befb392b3ab2cdfb77eed39155d7)

r-czajkowski (approved):

PR #905, opened 2024-11-27
Add retry logic to HTTP calls in SDK

nkuba:
We observed that the tBTC API calls are failing intermittently, which may be related to unstable Cloudflare workers infrastructure. This PR adds retry logic to the tBTC API calls to handle these failures. Requests to the tBTC API will be retried up to 5 times with an exponential backoff strategy.

Closes AENG-26


sdk/src/lib/api/HttpApi.ts
```diff
@@ -1,11 +1,48 @@
...
+    this.#retryOptions = retryOptions
+  }
+
+  /**
+   * Makes an HTTP request with retry logic.
+   * @param requestFn Function that returns a Promise of the HTTP response.
+   * @returns The HTTP response.
+   * @throws Error if the request fails after all retries.
+   */
+  async requestWithRetry(
```
r-czajkowski, line 31: What do you think about making this fn private and using it in `getRequest` and `postRequest`?

r-czajkowski, line 31: I think it can be `protected` or see comment above.

nkuba: I thought about it for a while. It may be worth to add retries by default.

nkuba: 2196fe39c1555a809a1668e9ca15813821f18528

r-czajkowski (approved):

PR #915, opened 2024-12-04
Update the directory structure in the dApp

kkosiorowska:
Closes AENG-158

This PR fixes the directory structure for dApp a bit. Let's make further possible tweaks in the next PRs.  What has been done:

- Reduced the directory structure and pull the existing components one level up.
- Used the same names on the file and the component inside. We should not decouple file name from component name—these should match.


dapp/src/components/TransactionModal/ActiveStakingStep/StakeFormModal/StakeDetails.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 import { List } from "@chakra-ui/react"
-import TransactionDetailsAmountItem from "#/components/shared/TransactionDetails/AmountItem"
-import FeesDetailsAmountItem from "#/components/shared/FeesDetails/FeesItem"
+import TransactionDetailsAmountItem from "#/components/shared/TransactionDetails/TransactionDetailsAmountItem"
+import FeesDetailsAmountItem from "#/components/shared/FeesDetails/FeesDetailsAmountItem"
```
nkuba, line 4: The file prefix duplicates the directory. Could we remove the prefix and differentiate import names?

```suggestion
import TransactionDetailsAmountItem from "#/components/shared/TransactionDetails/AmountItem"
import FeesDetailsAmountItem from "#/components/shared/FeesDetails/AmountItem"
```

r-czajkowski (dismissed):

dapp/src/components/TransactionModal/ActiveStakingStep/StakeFormModal/StakeDetails.tsx
kkosiorowska: I would leave it as is at this moment because this PR focuses on introducing the following two rules:

- Reduce the directory structure and pull the existing components one level up.
- Use the same names on the file and the component inside. We should not decouple file name from component name—these should match

Of course, the structure is still not perfect, but let's make the next improvements in the next PRs. I'm still in the process of creating tasks. Because there are still a few things to improve in the code regarding names or the directory structure.

In fact, from what I've seen `TransactionDetailsItem` and `FeesDetailsItem` and `FeesDetailsAmountItem` and `TransactionDetailsAmountItem` are virtually the same components that we could refactor and have generic list items. I would like to fix and names at that time.

PR #917, opened 2024-12-05
dApp: Wallet connection info alert

kpyszkowski:
Closes: AENG-24

This pull request focuses on refactoring the wallet connection error handling system to use alerts of multiple types. The changes include renaming components and contexts, updating imports and hooks, and modifying the structure of connection alerts.

### Changes: 
- Remove `status` from `WalletConnectionAlertContext`: Storing `status` in context is redundant. It was always `"error"` and having it in context made it difficult to control. The `status` property has only presentational purpose so it should be adjustable at the component level.
- Adjust `Alert` component styling: Change title font weight and removed margin.
- Adjust `Modal` component styling to match designs: Update paddings to match designs

<details>
  <summary>Before:</summary>
  image
</details>

<details>
  <summary>After:</summary>
  image
</details>


dapp/src/components/shared/Alert.tsx
```diff
@@ -22,7 +22,7 @@ import Spinner from "./Spinner"
 const STATUSES = {
   info: {
     icon: IconInfoCircle,
-    colorScheme: "gold",
+    colorScheme: "blue",
```
kkosiorowska, line 25: This change affects all alerts in the dApp. We probably don't want that but it should be confirmed with Sorin. We should probably set `colorScheme` when we define `CONNECTION_ALERTS`. This solution seems the easiest until we have a styleguide.

Screenshot 2024-12-10 at 09 07 11

Screenshot 2024-12-10 at 09 16 13

kpyszkowski: Fixed. Ref commit: 8f4faffa

image
image
image

kkosiorowska (approved):

PR #921, opened 2024-12-10
Move `.env` to `.env.example`

kkosiorowska:
Closes AENG-176

We should not commit the .env file, only .env.example should be committed.


dapp/.env.example
```diff
@@ -1,34 +1,32 @@
...
-VITE_FEATURE_FLAG_WITHDRAWALS_ENABLED="false"
-VITE_FEATURE_FLAG_OKX_WALLET_ENABLED="false"
-VITE_FEATURE_FLAG_XVERSE_WALLET_ENABLED="false"
+VITE_FEATURE_FLAG_WITHDRAWALS_ENABLED="true"
+VITE_FEATURE_FLAG_OKX_WALLET_ENABLED="true"
+VITE_FEATURE_FLAG_XVERSE_WALLET_ENABLED="true"
 VITE_FEATURE_FLAG_ACRE_POINTS_ENABLED="true"
 VITE_FEATURE_FLAG_TVL_ENABLED="true"
 VITE_FEATURE_GATING_DAPP_ENABLED="true"
 VITE_FEATURE_POSTHOG_ENABLED="true"
```
r-czajkowski, line 31: I would change it to `false` by default. We do not use the posthog on development.

dapp/.env.example
r-czajkowski, line 1: I believe we should add the `.env` file to `.gitignore`. We can add this:

```
.env.*
!.env.example
```

kpyszkowski: Yes please 🙏🏼 We have to

kkosiorowska: We do not need to do this because it is already added to `.gitignore`. In my case, the `.env` file is ignored by git.

https://github.com/thesis/acre/blob/3f127110c0a87c8a76e922190f9440dc456dd92f/.gitignore#L14-L19

dapp/.env.example
kkosiorowska: https://github.com/thesis/acre/pull/921/commits/795814037f0c2c5e35a0cf722e00231fec068ab5

r-czajkowski (approved):

PR #925, opened 2024-12-13
Update the Acre badge

kkosiorowska:
Closes AENG-170

We should not use the word staking, the new icon no longer uses it.

**Before**
Screenshot 2024-12-13 at 09 33 47

**After**

Screenshot 2024-12-13 at 09 33 33


dapp/src/components/MobileModeBanner/AcreAnimatedBadge.tsx
```diff
@@ -7,9 +7,10 @@ function AcreAnimatedBadge(props: BoxProps) {
     <Box
       as="svg"
       xmlns="http://www.w3.org/2000/svg"
-      w="175px"
-      h="173px"
+      w="224px"
```
kpyszkowski, line 10: Non blocking:
How about `rem`'s here?

kkosiorowska: I used px here to use the same value as in the `viewBox`. Actually, we pass the right size in props so I think we can get rid of these values. https://github.com/thesis/acre/pull/925/commits/618b5182e33c7144fccd77e0d6a0b56ab9270c65

kpyszkowski (approved):
LGTM

PR #943, opened 2025-05-08
Deposit fees

r-czajkowski:
WIP...

Some of the fees may be covered by Acre, so we want to display information about this in the tooltip.

<details>
  <summary>Screenshots</summary>

obraz

</details>

TODO:

- [ ] update the `estimateDepositFee` function in SDK. Take into account the reimbursement threshold from #942 and the tBTC balance of the `FeesReimbursementPool` contract.
- [ ] unit tests.

### DIFF
--- a/dapp/src/components/TransactionModal/ActiveUnstakingStep/SignMessageModal.tsx
+++ b/dapp/src/components/TransactionModal/ActiveUnstakingStep/SignMessageModal.tsx
@@ -129,7 +129,10 @@ export default function SignMessageModal() {
             status: "pending",
             // This is a requested amount. The amount of BTC received will be
             // around: `amount - transactionFee.total`.
-            amount: amount - transactionFee.acre,
+            // TODO: Based on the comment above: shouldn't we use total fee
+            // instead of only Acre fee. We can also use `estimatedAmount` from
+            // `useTransactionDetails` hook.
+            amount: amount - transactionFee.acre.fee,
             initializedAt: timeUtils.dateToUnixTimestamp(),
             // The message is signed immediately after the initialization.
             finalizedAt: timeUtils.dateToUnixTimestamp(),
--- a/dapp/src/hooks/useTransactionFee.ts
+++ b/dapp/src/hooks/useTransactionFee.ts
@@ -1,12 +1,13 @@
 import { useAcreContext } from "#/acre-react/hooks"
+import { Fees as ProtocolFees } from "@acre-btc/sdk"
 import { logPromiseFailure } from "#/utils"
 import { useEffect, useState } from "react"
 import { ACTION_FLOW_TYPES, ActionFlowType, Fee } from "#/types"
 import { useAppDispatch } from "./store"
 
-export const initialFee = {
-  tbtc: 0n,
-  acre: 0n,
+export const initialFee: Fee = {
+  tbtc: { fee: 0n, isReimbursable: false },
+  acre: { fee: 0n, isReimbursable: false },
   total: 0n,
 }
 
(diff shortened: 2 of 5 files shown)


dapp/src/components/TransactionModal/ActiveUnstakingStep/SignMessageModal.tsx
```diff
@@ -129,7 +129,10 @@ export default function SignMessageModal() {
             status: "pending",
             // This is a requested amount. The amount of BTC received will be
             // around: `amount - transactionFee.total`.
-            amount: amount - transactionFee.acre,
+            // TODO: Based on the comment above: shouldn't we use total fee
+            // instead of only Acre fee. We can also use `estimatedAmount` from
+            // `useTransactionDetails` hook.
+            amount: amount - transactionFee.acre.fee,
```
nkuba, line 135: Where is this value used?

dapp/src/hooks/useTransactionFee.ts
```diff
@@ -4,9 +4,9 @@ import { useEffect, useState } from "react"
 import { ACTION_FLOW_TYPES, ActionFlowType, Fee } from "#/types"
 import { useAppDispatch } from "./store"
 
-export const initialFee = {
-  tbtc: 0n,
-  acre: 0n,
+export const initialFee: Fee = {
```
nkuba, line 7:
```suggestion
export const initialFees: Fees = {
```

r-czajkowski: 87d3ea6

dapp/src/hooks/useTransactionDetails.ts
```diff
@@ -1,41 +1,15 @@
...
-      })
-    }
-  }, [amount, transactionFee])
+  const { data: transactionFee } = useTransactionFee(amount, flow)
 
-  return details
+  return {
+    amount,
+    transactionFee,
+    estimatedAmount: amount ?? 0n - transactionFee.total,
```
kpyszkowski, line 13: Do we allow `estimatedAmount` to be negative by subtracting from zero?

dapp/src/hooks/useTransactionFee.ts
```diff
@@ -1,44 +1,36 @@
...
-      const getEstimatedDepositFee = async () => {
-        if (!acre) return
+  return useQuery({
+    queryKey: [
+      ...queryKeysFactory.userKeys.estimateFee(),
+      flow,
+      amount?.toString(),
+    ],
+    queryFn: () => {
+      if (!acre || !amount) return initialFee
```
kpyszkowski, line 25: I'm not a fan of initial values in Tanstack Query.
It leads to problems with error handling - library doesn't know if error occured, data was missing or anything else.
I would rather throw an error and handle properly.

sdk/src/lib/ethereum/bitcoin-depositor.ts
```diff
@@ -226,13 +258,66 @@ class EthereumBitcoinDepositor
...
 
+  async #getTbtcBalanceOfFeesReimbursementPool() {
+    if (!this.#tbtcToken) throw new Error("tBTC contracts not set")
+
+    const feesReimbursementPool = await this.#getFeesReimbursementPool()
+
+    if (feesReimbursementPool.equals(EthereumAddress.from(ZeroAddress)))
+      return 0n
+
+    return this.#tbtcToken.balanceOf(await this.#getFeesReimbursementPool())
```
kpyszkowski, line 269: `feesReimbursementPool` can be reused

dapp/src/hooks/useTransactionDetails.ts
kpyszkowski: Also this hook seems to be redundant for now. It only proxies `useTransactionFee`.
It only transforms the shape of `data` and swallows errors at the same time 😆

dapp/src/hooks/useTransactionFee.ts
kpyszkowski: If you decide to add error handling, you could access `isError` flag by calling `useTransactionFee` in `Stake/UnstakeDetails` component and render alternative UI there.
It also justifies the redundancy of `useTransactionDetails` hook I mentioned.

Additionally worth considering implementing loading state by checking `isPending` flag.

If you check both of them, you don't have to worry about missing data (`initialData`). Also TypeScript will catch up with types - no `undefined` by type narrowing they have.

kpyszkowski (changes requested):

kpyszkowski (approved):

PR #947, opened 2025-07-02
Allocator contract for deposits into the Midas Vault

nkuba:
The MidasAllocator contract is a simple allocator that deposits tBTC to the
Midas Vault and holds the shares minted by the Vault.

The Vault is expected to be implemented by the Midas team according to the
IVault interface, to wrap their own vault contracts. As their setup uses two
distinct vaults for deposits and redemptions, in addition to a separate shares
token contract.

This implementation doesn't cover the withdrawal functionality, as it will
be built on top of the queued withdrawals functionality.

Closes: ACR-13


nkuba:
Closing in favor of https://github.com/acre-btc/acre/pull/954 that builds on top of these changes.

PR #957, opened 2025-07-27
Withdrawal status banner

r-czajkowski:
Closes: ACR-26

Implement the `WithdrawalStatusBanner` to display requested and ready-to-be-withdrawn withdrawals. As described in https://linear.app/acrefi/issue/ACR-23/asynchronous-withdrawals cancellation is out of the scope of this task so we do not render the `canel request` or `cancel` butons.

obraz

I've noted a few points where we need to decide whether to make changes or keep the current implementation:
- brown button variant - the button theme does not support the `colorScheme` so we can't easily render the brown button as on FIgma. Are we ok with outline button? If no we need to improve the button theme in a separate PR
   obraz
-  progress step component - same as above. We can't change the color of the progress.
   obraz


kpyszkowski (approved):
LGTM 🚀

PR #959, opened 2025-07-31
Make header "Connect wallet" button primary

cairomassimo:
Closes ACR-34


dapp/src/components/Header/ConnectWallet.tsx
```diff
@@ -67,9 +67,7 @@ export default function ConnectWallet() {
     return (
       <Button
         size="lg"
```
kpyszkowski, line 69: Changing variant changes also the size which is not desirable.
It should be addressed in the theme so `size` props controls the size and `variant` props controls the appearance.

Let's enforce the height of old button as a workaround with the proper TODO comment.

```suggestion
        size="lg"
        {/* TODO: Revisit Buttons theme and separate appearance (variant) from the size */}
        height="48px"
```

kpyszkowski (changes requested):
One remark

PR #963, opened 2025-08-25
Migrate stBTC to acreBTC

kpyszkowski:
Closes ACRE-65 ACRE-66 ACRE-67

### What's changed
- Updated dApp
- Updated SDK
  - Left TODOs for stuff to be done later (contracts related)
- Updated Subgraph
  - Left TODOs for stuff to be done later (contracts related)


r-czajkowski (approved):
LGTM! Should we wait for the new abi files before we merge this PR? I think it would be better to wait for the new contracts, just to make sure they have the same API.

PR #965, opened 2025-08-28
Update BitcoinDepositor to use acreBTC

nkuba:
BitcoinDepositor used to deposit funds in stBTC. Now it deposits in acreBTC.
The contract requires to be upgraded and the acreBTC has to be set on the
existing deployed instance.

Abstracted the destination contract to be defined as an interface to have
more flexibility with future upgrades.

Closes ACR-70


solidity/contracts/BitcoinDepositor.sol
```diff
@@ -193,12 +222,12 @@ contract BitcoinDepositor is AbstractTBTCDepositor, Ownable2StepUpgradeable {
...
         }
-        if (address(_stbtc) == address(0)) {
-            revert StbtcZeroAddress();
+        if (address(_acreVault) == address(0)) {
+            revert AcreVaultZeroAddress();
         }
 
         tbtcToken = IERC20(_tbtcToken);
-        stbtc = stBTC(_stbtc);
+        acreVault = IAcreVault(_acreVault);
```
pdyraga, line 230: I... don't think it's going to work. If we do an upgrade, we need a `initializeV2(...) reinitializer(2)` function.

nkuba: Yes, this line in the initializer is for fresh deployment. For the upgrade we won't run it but execute the update function as you noticed.

pdyraga (approved):

pdyraga (approved):

PR #968, opened 2025-09-01
WithdrawalQueue for Midas Vault

nkuba:
Depends on https://github.com/acre-btc/acre/pull/947
Replaces https://github.com/acre-btc/acre/pull/954

Midas AcreAdapter: https://github.com/midas-apps/contracts/pull/71

TODO:
- [x] improve unit tests
- [ ] update docs
- [ ] update PR description


solidity/contracts/acreBTC.sol
```diff
@@ -258,8 +303,8 @@ contract acreBTC is ERC4626Fees, PausableOwnable {
 
     /// @notice Withdraws assets from the vault and transfers them to the
     ///         receiver.
-    /// @dev Withdraw unallocated assets first and and if not enough, then pull
-    ///      the assets from the dispatcher.
+    /// @dev Withdraw unallocated assets first and if not enough, route through
+    ///      withdrawal queue.
```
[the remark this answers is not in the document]
nkuba: ddbcea3ddfce498e33dc604a8104f290594cbb43

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,163 @@
...
+
+import "@thesis-co/solidity-contracts/contracts/token/IReceiveApproval.sol";
+
+import "./stBTC.sol";
+import "./acreBTC.sol";
+import "./bridge/ITBTCToken.sol";
+import {ZeroAddress} from "./utils/Errors.sol";
+
+/// @title Bitcoin Redeemer
+/// @notice This contract facilitates redemption of stBTC tokens to Bitcoin through
```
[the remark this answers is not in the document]
nkuba: fea3bc243fb56ae3ce1ce877d94fdb14be94b4f1

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+/// @notice This contract facilitates redemption of acreBTC tokens to Bitcoin through
+///         tBTC redemption process.
+contract BitcoinRedeemerV2 is Ownable2StepUpgradeable, IReceiveApproval {
+    using SafeERC20 for IERC20;
+
+    /// Interface for tBTC token contract.
+    ITBTCToken public tbtcToken;
+
+    /// Address of the TBTCVault contract.
+    address public tbtcVault;
```
r-czajkowski, line 22: Do we need this? Looks like this is unnecessary.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Address of the Acre Vault contract.
+    acreBTC public acreBtc;
+
+    /// Emitted when the TBTCVault contract address is updated.
+    /// @param oldTbtcVault Address of the old TBTCVault contract.
+    /// @param newTbtcVault Address of the new TBTCVault contract.
+    event TbtcVaultUpdated(address oldTbtcVault, address newTbtcVault);
```
r-czajkowski, line 30: Same as above.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+        address indexed owner,
+        uint256 indexed requestId,
+        uint256 shares
+    );
+
+    /// Reverts if the tBTC Token address is zero.
+    error TbtcTokenZeroAddress();
+
+    /// Reverts if the TBTCVault address is zero.
+    error TbtcVaultZeroAddress();
```
r-czajkowski, line 46: Same as above.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+    error AcreBtcZeroAddress();
+
+    /// Attempted to call receiveApproval by supported token.
+    error CallerNotAllowed(address caller);
+
+    /// Attempted to call receiveApproval with empty data.
+    error EmptyExtraData();
+
+    /// Attempted to call _redeemSharesAndUnmint with unexpected tBTC token owner.
+    error UnexpectedTbtcTokenOwner();
```
r-czajkowski, line 58: Same as above.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Reverts if the redeemer is not the deposit owner.
+    error RedeemerNotOwner(address redeemer, address owner);
+
+    /// Reverts when approveAndCall to tBTC contract fails.
+    error ApproveAndCallFailed();
```
r-czajkowski, line 64: Let's remove it.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Reverts if the new TBTCVault contract is not tBTC token owner.
+    error NotTbtcTokenOwner();
```
r-czajkowski, line 67: Let's remove it.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+    ) public initializer {
+        __Ownable2Step_init();
+        __Ownable_init(msg.sender);
+
+        if (address(_tbtcToken) == address(0)) {
+            revert TbtcTokenZeroAddress();
+        }
+        if (address(_tbtcVault) == address(0)) {
+            revert TbtcVaultZeroAddress();
+        }
```
r-czajkowski, line 91: Let's remove it.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+        if (address(_acreBtc) == address(0)) {
+            revert AcreBtcZeroAddress();
```
r-czajkowski, line 93: Maybe we can use already defined `ZeroAddress` in `./utils/Errors.sol`

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        tbtcToken = ITBTCToken(_tbtcToken);
+        tbtcVault = _tbtcVault;
```
r-czajkowski, line 97: Let's remove it.

solidity/contracts/BitcoinRedeemerV2.sol
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        if (newTbtcVault != tbtcToken.owner()) {
+            revert NotTbtcTokenOwner();
+        }
+
+        emit TbtcVaultUpdated(tbtcVault, newTbtcVault);
+
+        tbtcVault = newTbtcVault;
+    }
```
r-czajkowski, line 133: Do we need this? It seems unnecessary to me.

solidity/contracts/acreBTC.sol
```diff
@@ -256,47 +303,94 @@ contract acreBTC is ERC4626Fees, PausableOwnable {
...
+    /// @notice Redeems shares for Bitcoin bridge withdrawal.
+    /// @dev This function routes through the withdrawal queue for bridge redemptions.
     /// @param shares Amount of shares to redeem.
-    /// @param receiver The address to which the assets will be transferred.
-    /// @param owner The address of the owner of the shares.
-    function redeem(
+    /// @param redeemerOutputScript Redeemer output script.
+    /// @return requestId The ID of the withdrawal request in the queue.
+    function requestRedeemAndBridge(
         uint256 shares,
```
pdyraga, line 365: There is no guarantee the provided shares amount represents enough tBTC so that requesting to-BTC redemption on the bridge is possible.

solidity/contracts/acreBTC.sol
```diff
@@ -183,6 +218,18 @@ contract acreBTC is ERC4626Fees, PausableOwnable {
...
+    /// @notice Updates the withdrawal queue contract address.
+    /// @param newWithdrawalQueue Address of the new withdrawal queue contract.
+    function updateWithdrawalQueue(
+        address newWithdrawalQueue
+    ) external onlyOwner {
+        // newWithdrawalQueue can be zero to disable the withdrawal queue
+        address oldWithdrawalQueue = withdrawalQueue;
+        withdrawalQueue = newWithdrawalQueue;
+
+        emit WithdrawalQueueUpdated(oldWithdrawalQueue, newWithdrawalQueue);
```
pdyraga, line 230: Or just emit it before assigning the new value. 😉

solidity/contracts/midas/MidasAllocator.sol
```diff
@@ -104,7 +116,18 @@ contract MidasAllocator is IDispatcher, Maintainable {
...
+        vaultSharesToken.forceApprove(address(midasVault), shares);
+        return midasVault.requestRedeem(shares, address(acreVault));
+    }
+
+    /// @notice Sets the withdrawal queue address.
+    /// @param _withdrawalQueue Address of the withdrawal queue.
+    function setWithdrawalQueue(address _withdrawalQueue) external onlyOwner {
+        if (_withdrawalQueue == address(0)) {
+            revert ZeroAddress();
+        }
```
pdyraga, line 130: In acreBTC, the withdrawal queue can be zero address to disable withdrawals. Why not allowing it to be zero here as well, for consistency? The transfer in `withdrawShares` is not going to work as a zero address can't call it.

solidity/contracts/BitcoinDepositorV2.sol
```diff
@@ -60,25 +95,22 @@ contract BitcoinDepositorV2 is AbstractTBTCDepositor, Ownable2StepUpgradeable {
     /// @notice Fees reimbursement pool.
     FeesReimbursementPool public feesReimbursementPool;
```
pdyraga, line 96: I have serious concerns about the reimbursement mechanism used in this contract.

The `tbtcAmount` returned from `_finalizeDeposit(depositKey)` is always lower than the `initialAmount` so we'll reimburse even if the amount of minted tBTC is equal to the `initialAmount`. As a result, we'll keep moving funds from `FeesReimbursementPool` to `BitcoinDepositorV2` with each individual deposit.

The approximation done by `_finalizeDeposit(depositKey)` is the worst-case scenario, and the worst-case scenario is far from reality, given that the [`depositTxMaxFee`](https://github.com/threshold-network/tbtc-v2/blob/49f82f5425b840b51e53a5cfbfdf464ddabf40a3/solidity/contracts/bridge/Bridge.sol#L1690-L1693) must be sufficiently high.

Some time ago, I analysed all BTC deposit transactions to Mezo and all of them were handled by tBTC optimistic minting. Even if they were not done by optimistic minting, the average sweep cost per deposit was $0.37. The `depositTxMaxFee` is currently set to 0.001 BTC, so $113 with today's price. The difference is significant.

It is impossible to say whether tBTC was minted via optimistic minting or sweeping. It's also not possible to distinguish individual tBTC minting operations to the `BitcoinDepositorV2` contract with concurrent deposits.

pdyraga: First, we want to ensure all tBTC deposits are done via optimistic minting so that we never have to deal with the sweep fee. This was the case for all Mezo deposits for the last year I've analyzed. If we want to be extra-sure, we can propose running a tBTC minter as Acre to the Threshold DAO.

Second, let's not look at `tbtcAmount` approximation from `_finalizeDeposit(depositKey)` at all, as it's too pessimistic and far from reality. Instead, let's check if the deposit was finalized by tBTC Bridge (we do that already) and if there is enough balance in the `BitcoinDepositorV2` contract to push it to `acreVault`. If not, let's draw from the reimbursement pool.

I would also completely eliminate `bridgeFeesReimbursementThreshold` - tBTC minted from concurrent deposits to the depositor contract is fungible. We need to draw from the reimbursement pool for all deposits, if needed.

This way, we will only need to draw from the reimbursement pool if optimistic minting in tBTC stops working, and even if we have to draw from there, the amounts will be around $0.37 per deposit and not around $113 per deposit.

solidity/contracts/midas/WithdrawalQueue.sol
```diff
@@ -0,0 +1,432 @@
...
+    /// @notice Updates the TBTCVault contract address.
+    /// @param newTbtcVault The new TBTCVault contract address.
+    function updateTbtcVault(address newTbtcVault) external onlyOwner {
+        if (newTbtcVault == address(0)) {
+            revert ZeroAddress();
+        }
+
+        if (newTbtcVault != tbtc.owner()) {
+            revert NotTbtcTokenOwner();
+        }
```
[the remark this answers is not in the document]
pdyraga: Good digging! This had more sense in #335, where it was used in `_redeemSharesAndUnmint`, and when to-BTC redeems were synchronous operations. In the current circumstances and with this code, I would argue it's just a waste of gas. Having this condition pass during the governance TX doesn't mean it will pass again during the redemption.

pdyraga (dismissed):

pdyraga (approved):

solidity/contracts/BitcoinDepositorV2.sol
pdyraga: Done in https://github.com/acre-btc/acre/pull/979.

PR #969, opened 2025-09-02
Vaults TVL cap

kpyszkowski:
Closes ACR-64

### Changes
- Adds TVL to vaults 
image


dapp/src/utils/numbersUtils.ts
```diff
@@ -218,23 +218,49 @@ const addLeadingZero = (num: number): string =>
 const getPercentValue = (value: number, maxValue: number) =>
   (value * 100) / maxValue
 
-type DesiredDecimals = Intl.NumberFormatOptions &
-  BigIntToLocaleStringOptions["minimumFractionDigits"]
+type FormatNumberToCompactStringOptions = {
+  decimals?: Intl.NumberFormatOptions &
+    BigIntToLocaleStringOptions["maximumFractionDigits"]
```
r-czajkowski, line 223: Why  is `decimals` typed as `Intl.NumberFormatOptions`? Should we define the  `FormatNumberToCompactStringOptions` type in this way:
```ts
type FormatNumberToCompactStringOptions = {
 decimals?: BigIntToLocaleStringOptions["maximumFractionDigits"],
  withVariableDecimals?: boolean
  currency?: "USD"
} & Intl.NumberFormatOptions
```

dapp/src/utils/numbersUtils.ts
```diff
@@ -218,23 +218,49 @@ const addLeadingZero = (num: number): string =>
...
+  withVariableDecimals?: boolean
```
r-czajkowski, line 224: I think we should rename this variable. It took me a while to understand what it actually does.

Maybe like this 🤔:
```suggestion
  withAutoCompactFormat?: boolean
```

Ideally, the variable name should suggest that the number is formatted based on predefined rules.

kpyszkowski: It is always a compact format. That's the purpose of the function.
The options is called `withVariableDecimals` because number of decimals might vary under specific conditions - on the number scale in this case:
- for millions only one decimal is presented
- for thousands - two

With this option the rule described above applies.
If not, the number of decimals will be constant - 2 if currency, 4 otherwise, or anything else if `decimals` is explicitly provided.

kpyszkowski: Gave more detailed explainations in JSDoc

be6d037d

dapp/src/utils/numbersUtils.ts
r-czajkowski: > I don't want that much flexibility. Constraints are clear:
> 
>     * compact numbers
> 
>     * complete flexibility on decimals
> 
>     * additional currency formatting
> 
> 
> This type definition is too general - accepts too much options which won't be applied. Wrapping with really long `Omit<>` or `Pick<>` is not an option.

I understand the constraint but the type definition didn’t reflect it.

> I shouldn't make union to Intl.NumberFormatOptions

Yeah exactly this is what I meant.

dapp/src/utils/numbersUtils.ts
r-czajkowski: > It is always a compact format. That's the purpose of the function.

Yes I understand the purpose of the function. The only issue I see is the misleading name of the option. That's why I added `Auto` (`withAutoCompactFormat`) (or maybe I should use the `default` word) to indicate that the number is formatted according to the default rules.

At first glance, the `withVariableDecimals` suggests that the same number of decimal places as the variable currently has will be used, instead of being reformatted using the default rules.

dapp/src/utils/numbersUtils.ts
kpyszkowski: Sure. So feel free to resolve the conversation if my changes work for you 🙏🏼

dapp/src/utils/numbersUtils.ts
kpyszkowski: 95c2315e

PR #1014, opened 2025-11-03
Fetch fees percentages from SDK to show in vault details

nkuba:
The vault details have hardcoded values for the fees percentages. This commit fetches the fees percentages from the SDK to show in the vault details.

We also show TVL values in the vault details.

### DIFF
--- a/dapp/src/components/MidasVaultDetails.tsx
+++ b/dapp/src/components/MidasVaultDetails.tsx
@@ -1,3 +1,4 @@
+import React from "react"
 import TbtcIcon from "#/assets/icons/TbtcIcon"
 import { externalHref, vaults } from "#/constants"
 import { addressUtils } from "#/utils"
--- a/dapp/src/components/Vaults.tsx
+++ b/dapp/src/components/Vaults.tsx
@@ -1,3 +1,4 @@
+import React from "react"
 import { vaults } from "#/constants"
 import {
   useCurrencyConversion,


dapp/src/hooks/useWithdrawalFeePercentage.ts
```diff
@@ -0,0 +1,30 @@
...
+      )
+
+      const feePercentage =
+        (Number(fees.total) / Number(SAMPLE_DEPOSIT_AMOUNT)) * 100
+      return feePercentage
+    },
+    enabled: !!acre,
+    // Cache the result for 5 minutes since fee percentages don't change frequently
+    staleTime: 5 * 60 * 1000,
+  })
```
r-czajkowski: Let's use `useTransactionFee()` hook:
```ts
export default function useWithdrawalFeePercentage(flow: ActionFlowType) {
  const { data: fees, ...rest } = useTransactionFee(SAMPLE_AMOUNT, flow, {
    staleTime: time.ONE_MINUTE_IN_SECONDS * 5,
    queryKey: [
      ...queryKeysFactory.userKeys.estimateFee(),
      "withdrawalFeePercentage",
    ],
  })

  const data = useMemo(() => {
    if (!fees) return 0

    return (Number(fees) / Number(SAMPLE_DEPOSIT_AMOUNT)) * 100
  }, [fees])

  return { data, ...rest }
}
```

dapp/src/hooks/useDepositFeePercentage.ts
```diff
@@ -0,0 +1,28 @@
+import { useAcreContext } from "#/acre-react/hooks"
+import { queryKeysFactory } from "#/constants"
+import { useQuery } from "@tanstack/react-query"
+
+const SAMPLE_DEPOSIT_AMOUNT = 100000000n // 1 BTC in satoshis for fee calculation
+
+export default function useDepositFeePercentage() {
```
r-czajkowski: Same as above, we can have on hook and pass just flow type.

dapp/src/hooks/useWithdrawalFeePercentage.ts
nkuba: I'll address it in a follow-up PR

r-czajkowski (approved):

PR #1042, opened 2026-08-31
Withdraw to tBTC on Ethereum

r-czajkowski:
Adds a second withdrawal destination: redeem an acreBTC position for tBTC paid out to an Ethereum address, instead of bridging to Bitcoin. 

Goes through `acreBTC.requestRedeem` → `WithdrawalQueue.requestRedeem` → Midas. `WithdrawalQueue.requestRedeem` is `onlyAcreBTC`, so acreBTC is the only possible entrypoint. Like the Bitcoin path it is asynchronous — the transaction registers the request, and the tBTC arrives when the Midas redemption settles at the next NAV update. It skips the tBTC Bridge, so the Bridge redemption fee and the dust-threshold minimum do not apply.
  
  **SDK** — `Account#initializeTbtcWithdrawal`, plus `requestRedeem` encoding and `RedemptionRequested` lookup on the acreBTC handle. The signing callback is shared with `initializeWithdrawal`.

  **dapp** — a checkbox and address field on the withdraw form, the destination tracked in the action-flow store, and fees quoted per destination. Bitcoin remains the default.

**Known gap:** the subgraph indexes only `RedeemAndBridgeRequested`, so a tBTC withdrawal shows optimistically and then disappears from the dashboard on the next refetch. Indexing is tracked separately.

NOTE: withdrawals are still behind VITE_FEATURE_FLAG_WITHDRAWALS_ENABLED, so we need to enable this feature flag before shipping to prod.


dapp/src/utils/forms.ts
```diff
@@ -56,6 +57,41 @@ function validateTokenAmount(
...
+  // viem's default `strict: true` would reject the uppercase one. Genuinely
+  // corrupted mixed-case addresses are still caught downstream, where the SDK
+  // runs them through `EthereumAddress.from`.
+  if (!isAddress(address, { strict: false })) return ERRORS.INVALID
+
+  // `isAddressEqual` throws on a malformed argument, so the forbidden address
+  // is shape-checked first - it comes from the connected wallet and may be
+  // absent before the account resolves.
+  if (
+    options?.forbiddenAddress &&
```
jagodarybacka, line 100: where is the check for zero address?

dapp/src/hooks/useWithdrawalFeePercentage.ts
```diff
@@ -1,22 +1,28 @@
...
 import { queryKeysFactory } from "#/constants"
+import { WithdrawalDestination } from "#/types"
+import { feesUtils } from "#/utils"
 import { useQuery } from "@tanstack/react-query"
 
 const SAMPLE_DEPOSIT_AMOUNT = 100000000n // 1 BTC in satoshis for fee calculation
 
-export default function useWithdrawalFeePercentage() {
+export default function useWithdrawalFeePercentage(
+  withdrawalDestination: WithdrawalDestination["type"] = "bitcoin",
```
jagodarybacka, line 10: seems like [here](https://github.com/acre-btc/acre/blob/withdraw-to-tbtc-from-midas/dapp/src/components/Vaults.tsx#L93) we are not passing the `withdrawalDestination` - is this correct?

dapp/src/components/TransactionModal/ActiveUnstakingStep/UnstakeFormModal/UnstakeFormBase.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <FormTokenBalanceInput
+        name={TOKEN_AMOUNT_FIELD_NAME}
+        tokenBalance={tokenBalance}
+        placeholder={tokenBalanceInputPlaceholder}
+        tokenAmountLabel={tokenAmountLabel}
+        currency={currency}
+        // A withdrawal always exits the whole position, so the amount is fixed
+        // to the balance and the field is read-only.
+        defaultAmount={tokenBalance}
+        isDisabled
```
jagodarybacka: So now user can no longer do partial withdrawals, no matter the destination?

jagodarybacka (commented):
A couple more questions:

1. I can't manually test this path but I wonder which success modal we are using - [this one](https://github.com/acre-btc/acre/blob/withdraw-to-tbtc-from-midas/dapp/src/components/TransactionModal/SuccessModal.tsx#L70)? If so then the  text on the success modal is not correct regarding the time user will receive funds, no?
2. Regarding withdrawals history and missing withdrawals to tbtc - don't we want to add local memory of these withdrawals if we don't want to add another subgraph?
3. The alerts regarding the withdrawal time are still visible for tbtc path even tho these are not the same as on bitcoin 

Image

dapp/src/components/TransactionModal/ActiveUnstakingStep/UnstakeFormModal/UnstakeFormBase.tsx
r-czajkowski: Yes. We want users to withdraw all their funds.

dapp/src/hooks/useWithdrawalFeePercentage.ts
r-czajkowski: Yeah I think so. We're focused on Bitcoin-native users, so it should stay as is imo. In the withdrawal form we display fees based on the destination.

dapp/src/utils/forms.ts
r-czajkowski: [f23c377](https://github.com/acre-btc/acre/pull/1042/commits/f23c37727ec5031826945c46eccc7572fa04f556)

nkuba (approved):

PR #1045, opened 2026-09-08
Track tBTC withdrawals in subgrpah

r-czajkowski:
Until now the subgraph only indexed withdrawals that go to Bitcoin. A withdrawal paid out in tBTC showed up in the dapp right after signing and then disappeared on the next refetch, because nothing was indexing it. There are already 23 of them on mainnet.
  
### Subgraph
- Index both tBTC payout paths: a queued redemption (acreBTC.RedemptionRequested) and a direct redeem when the queue is unset (the ERC4626 Withdraw event). Amounts come from the queue's RedeemRequested / RedeemFeeRequested.
- A queued redemption is only complete once Midas settles, and nothing on the Acre side fires then, so completion comes from the Midas redemption vault's SafeApproveRequest. The new MidasRequestToWithdrawal entity remembers which withdrawal filed which Midas request.
- Index BitcoinRedeemerV3. It is not queued, so it also emits Withdraw, which would otherwise be recorded as a finished tBTC withdrawal. It is now recognised as a Bitcoin withdrawal instead.
- Withdraw gains a destination field (Bitcoin or Ethereum), defaulting to Bitcoin so nothing existing changes.
- tbtcBridge gets its own startBlock instead of borrowing the redeemer's, so repointing a redeemer can no longer truncate the Bridge's indexing range.

### SDK
- Pass the destination through, and read withdrawal events by type instead of by position. tBTC withdrawals never touch the tBTC Bridge, so they have no Initialized stage and positional reads made every settled one look pending.

### dapp
- Withdrawals to tBTC show a tBTC on Ethereum tag and link to Etherscan. Before, every withdrawal looked the same and the Details link always pointed at a Bitcoin explorer, which resolves to nothing for a tBTC withdrawal.

- The withdraw amount field is locked to the whole position. It was already seeded with the full balance and a withdrawal exits the whole position, so any other amount the user typed could not be acted on.

### Deployment
The subgraphs are deployed on both mainnet and testnet (Sepolia), with the `MidasRedemptionVault` and `BitcoinRedeemerV3` addresses filled in.

### DIFF
--- a/sdk/src/lib/api/AcreSubgraphApi.ts
+++ b/sdk/src/lib/api/AcreSubgraphApi.ts
@@ -76,6 +76,14 @@ type Deposit = {
   finalizedAt?: number
 }
 
+/**
+ * Where the redeemed assets are delivered. `Bitcoin` goes through the tBTC
+ * Bridge, `Ethereum` pays tBTC out to an Ethereum address.
+ */
+export type WithdrawalDestination = "bitcoin" | "ethereum"
+
+type WithdrawalEventType = "Requested" | "Initialized" | "Finalized"
+
 type WithdrawalsDataResponse = {
   data: {
     withdraws: {
@@ -84,8 +92,10 @@ type WithdrawalsDataResponse = {
       amount: string
       requestedAmount: string
       amountToRedeem: string
+      destination: "Bitcoin" | "Ethereum"
       events: {
-        type: "Requested" | "Initialized" | "Finalized"
+        id: string
+        type: WithdrawalEventType
         timestamp: string
       }[]
     }[]
@@ -116,6 +126,14 @@ type Withdrawal = {
    */
   bitcoinTransactionId?: string
   /**
+   * Where the withdrawal is delivered.
+   */
+  destination: WithdrawalDestination
+  /**
+   * Hash of the Ethereum transaction that requested the withdrawal.
+   */
+  ethereumTransactionId: string
+  /**
    * Timestamp when the withdrawal was requested.
    */
   requestedAt: number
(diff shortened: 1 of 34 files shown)


sdk/src/lib/api/AcreSubgraphApi.ts
```diff
@@ -261,23 +295,34 @@ export default class AcreSubgraphApi extends HttpApi {
...
+        destination:
+          withdraw.destination === "Ethereum"
+            ? ("ethereum" as const)
+            : ("bitcoin" as const),
+        ethereumTransactionId: requestedEvent
+          ? getTransactionHashFromEventId(requestedEvent.id)
+          : "",
+        requestedAt: requestedEvent
+          ? parseInt(requestedEvent.timestamp, 10)
+          : 0,
```
michalsmiarowski, line 323: When there is no `Requested` event we return `""` and `0`. The type says `ethereumTransactionId: string` ("Hash of the Ethereum transaction"), so the type is lying — and `requestedAt: 0` makes the row sort to the very bottom of the history with the date 1970, with nothing in the logs.

Maybe we could show that the value is missing, instead of sending an empty string?

```ts
ethereumTransactionId?: string   // in both Withdrawal types
```

```ts
ethereumTransactionId: requestedEvent
  ? getTransactionHashFromEventId(requestedEvent.id)
  : undefined,
```

The dapp already does `activity.txHash ? ... : <Box/>`, so `undefined` works there with no change. For `requestedAt` I would keep the `0` but add a log, so we notice the bad row instead of silently hiding it at the bottom of the list.

r-czajkowski: I took a look at this case and imo it's not a real case. There must be something wrong with the indexer, or we may have set the starting blocks of the contracts incorrectly.

A withdrawal ends up with no `Requested` event only if the subgraph indexes its **ending** but misses its **beginning**. That needs the request to happen before the `startBlock` of the contract that records it.

michalsmiarowski (approved):
LGTM

pdyraga (approved):

